PR #2, opened 2020-02-20
Feature: Test Orchestrator (astrolabe)

prashantmital:
Still needs:

- [x] CLI endpoint to run a single spec test file
- [x] ~~mongod/mongos log retrieval functionality (wait until Atlas logs are current for driver ops; download them in a folder hierarchy that makes it easy to find the logs corresponding to a specific test; zip this file for upload to S3 after EVG run)~~ (moved to follow-on work, see #4)
- [x] cluster cleanup functionality (delete clusters once a test run is complete and logs for it have been downloaded to the EVG host)
- [x] Logging using Python's stdlib `logging` module (this includes INFO level logs instead of the current print statements that make it easier to follow the progress of a test run)


prashantmital:
These changes are based on #1. Merging that first should make reviewing this a whole lot easier.

astrolabe/commands.py
```diff
@@ -0,0 +1,138 @@
...
+    project = client.groups.byName[group_name].get().data
+    cluster = client.groups[project.id].clusters[cluster_name].get().data
+    return cluster.stateName == target_state
+
+
+def wait_until_cluster_state(client, group_name, cluster_name, target_state,
+                             polling_frequency, polling_timeout):
+    if is_cluster_state(client, group_name, cluster_name,
+                        target_state):
+        return True
```
ShaneHarvey: This seems redundant given the same call happens below. Do we need this check?

astrolabe/commands.py
```diff
@@ -0,0 +1,138 @@
...
+    sleep_interval = 1 / polling_frequency
+    while (time() - start_time) < polling_timeout:
+        if is_cluster_state(client, group_name, cluster_name, target_state):
+            return True
+        sleep(sleep_interval)
+
+    return False
+
+
+def select_callback(callback, args, kwargs, frequency, timeout):
```
ShaneHarvey: It seems like select_callback, wait_until_cluster_state, and BooleanCallableSelector all call a method every X seconds until that method returns true. Can they all use the same loop?

setup.py
```diff
@@ -40,6 +40,7 @@
         'pymongo>=3.10,<4',
         'dnspython>=1.16,<2',
         'pyyaml>=5,<6',
+        'tabulate>=0.8,<0.9',
```
ShaneHarvey: What feature does this library provide? Do we need it?

atlasclient/utils.py
```diff
@@ -21,12 +21,20 @@
...
         if name in self:
-            return self[name]
-        raise AttributeError('{} has no property named {}}.'.format(
+            return self.__coerce_dict(self[name])
+        raise AttributeError('{} has no property named {}.'.format(
             self.__class__.__name__, name))
 
+    def __getitem__(self, item):
+        return self.__coerce_dict(super().__getitem__(item))
+
```
ShaneHarvey: Why do we need these changes?

astrolabe/cli.py
```diff
@@ -0,0 +1,454 @@
...
+@click.pass_context
+def check_connection(ctx):
+    """Command to verify validity of Atlas API credentials."""
+    pprint(ctx.obj.root.get().data)
+
+#
+# @cli.command()
+# @click.option('-c', '--config', multiple=True, type=_JsonDotNotationType())
+# def cluster_config_option(config):
+#     pprint(_merge_dictionaries(config))
```
ShaneHarvey: Is this needed?

astrolabe/commands.py
```diff
@@ -0,0 +1,138 @@
...
+        else:
+            raise
+
+
+def ensure_connect_from_anywhere(*, client, group_id,):
+    ip_details_list = [{"cidrBlock": "0.0.0.0/0"}]
+    client.groups[group_id].whitelist.post(json=ip_details_list)
+
+
+def get_cluster_state(client, group_name, cluster_name):
```
ShaneHarvey: Why do some functions have `*` and some don't?

astrolabe/utils.py
```diff
@@ -0,0 +1,132 @@
...
+def _merge_dictionaries(dicts):
+    """Utility method to merge a list of dictionaries.
+    Last observed value prevails."""
+    result = {}
+    for d in dicts:
+        result.update(d)
+    return result
+
+
+class _JsonDotNotationType(click.ParamType):
```
ShaneHarvey: A unit test for this would be nice. Or at least some sample input/output in the docstring.

atlasclient/utils.py
prashantmital: This allows dot-access to nested dictionary members which is used quite often. With the older code only the first-level keys of the dictionary could be accessed via the `obj.attr` pattern. This change ensures that this kind of access is possible for all nested dicts as well.

setup.py
prashantmital: It provides a quick way to convert a list of lists into a textual table of various styles. In our case, we want the rst format to be able to include the resulting table into docs. While the docs part is not yet done, take a look at the `docgen.py` file for a closer look.
To see it in action install the package and do `astrolabe info environment-variables` or ` astrolabe info default-values`.

astrolabe/commands.py
prashantmital: This was old code. Please disregard and take another look. The "selector" is now only implemented in `poller.py` (might be in need of a name change).

astrolabe/commands.py
prashantmital: Old code. Please disregard.

astrolabe/cli.py
prashantmital: Removed in the latest commit.

astrolabe/commands.py
prashantmital: This was old code and is now removed. For the most part, I have tried to have functions be kwarg only, unless doing so would make things exceptionally verbose. I find that doing this improves the readability of code and reduces chances of errors.

astrolabe/utils.py
prashantmital: I can add a unittest but I am not so sure we need to support free-form text input of JSON-fields (which is what this class was designed for) anymore which might mean we can get rid of this altogether. Fleshing out the last remaining CLI endpoints will make things clearer.

prashantmital:
Sorry for the epic code churn in that last commit @ShaneHarvey. The next time you take a look, things should make a lot more sense.

astrolabe/utils.py
prashantmital: This is no longer needed.

astrolabe/cli.py
```diff
@@ -0,0 +1,484 @@
...
+from astrolabe.spec_runner import MultiTestRunner, SingleTestRunner
+from astrolabe.configuration import (
+    CLI_OPTION_NAMES as OPTNAMES,
+    CONFIG_DEFAULTS as DEFAULTS,
+    CONFIG_ENVVARS as ENVVARS,
+    TestCaseConfiguration)
+from astrolabe.utils import ClickLogHandler
+
+
+__logger__ = logging.getLogger(__name__)
```
ShaneHarvey: Why `__logger__` and not simply `logger`?

astrolabe/spec_runner.py
```diff
@@ -0,0 +1,403 @@
...
+            __logger__.info(
+                "Successfully loaded test data on cluster {!r}".format(
+                    self.cluster_name))
+
+        # Step-2: run driver workload.
+        __logger__.info("Starting workload executor")
+        connection_string = self.get_connection_string()
+        driver_workload = json.dumps(self.spec.driverWorkload)
+        worker_subprocess = subprocess.Popen([
+            sys.executable, self.config.workload_executor, connection_string,
```
ShaneHarvey: Does `sys.executable` belong here?

astrolabe/spec_runner.py
```diff
@@ -0,0 +1,403 @@
...
+            driver_workload], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
ShaneHarvey: Since we PIPE the output, we need to continuously read the worker_subprocess output (via `worker_subprocess.stdout.read(..)`) or else we risk blocking the child process when the pipe fills up.

See the note in subprocess.wait:
> This will deadlock when using stdout=PIPE or stderr=PIPE and the child process generates enough output to a pipe such that it blocks waiting for the OS pipe buffer to accept more data. Use Popen.communicate() when using pipes to avoid that.

https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait

A simpler option might be to redirect the worker_subprocess output to a file. Then we can open and read the file after the process exits.

ShaneHarvey: Just to be clear: there's no risk of deadlock in the current approach. The risk is that the child process can fill up the pipe with output and then it will block until we call worker_subprocess.communicate() down below.

astrolabe/cli.py
prashantmital: changed to LOGGER

astrolabe/spec_runner.py
prashantmital: Absolutely not. Removed.

astrolabe/spec_runner.py
prashantmital: Great point. I have opened #9 for this.

astrolabe/spec_runner.py
```diff
@@ -0,0 +1,400 @@
...
+        test_data = self.spec.driverWorkload.get('testData')
+        if test_data:
+            LOGGER.info("Loading test data on cluster {!r}".format(
+                self.cluster_name))
+            connection_string = self.get_connection_string()
+            client = MongoClient(connection_string, w="majority")
+            coll = client.get_database(
+                self.spec.driverWorkload.database).get_collection(
+                self.spec.driverWorkload.collection)
+            coll.drop()
```
mbroadst: do you need to potentially catch `ns not found`? or is that not a problem for python?

astrolabe/spec_runner.py
```diff
@@ -0,0 +1,400 @@
...
+            driver_workload], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        LOGGER.info("Started workload executor [PID: {}]".format(
+            worker_subprocess.pid))
+
+        # Step-3: begin maintenance routine.
+        final_config = self.spec.maintenancePlan.final
+        cluster_config = final_config.clusterConfiguration
+        process_args = final_config.processArgs
+
+        if not cluster_config and not process_args:
```
mbroadst: should we consider doing validation much earlier, so you can bail before any I/O if need be?

astrolabe/cli.py
```diff
@@ -66,6 +66,20 @@
...
+        exists=True, file_okay=True, dir_okay=False, resolve_path=True),
+    help='Absolute or relative path to the workload-executor')
+
+CLUSTERNAMESALT_OPTION = click.option(
+    OPTNAMES.CLUSTER_NAME_SALT, type=click.STRING, required=True,
+    envvar=ENVVARS.CLUSTER_NAME_SALT,
+    help='Salt for generating unique hashes.')
+
+XUNITOUTPUT_OPTION = click.option(
+    '--xunit-output', type=click.STRING, default="xunit-output",
```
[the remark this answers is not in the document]
prashantmital: The ones which can be displayed/tabulated using `astrolabe info <...>` are the ones that are specified using a `OPTNAMES.XXX`. Options with inline strings are not tabulated in the same manner and are not required to be (in my estimation).

astrolabe/spec_runner.py
prashantmital: PyMongo ignores this error on `drop()` so no special handling is required.

astrolabe/spec_runner.py
prashantmital: Good point. In general, we just need a lot more validation of the test specification. I have opened https://github.com/mongodb-labs/drivers-atlas-testing/issues/10 to do this.

PR #3, opened 2020-02-24
atlasclient improvements

prashantmital:
**Please review this before #2** 

In this PR:

1. JSONObject has been refactored to return nested dictionaries as JSONObject instances instead of vanilla dictionaries. This ensures we can continue to use the `object.attribute` access pattern to retrieve nested entries in JSON objects.
2. Tests have been added (`atlasclient/tests/`) to test the functionality introduced above. An entry has been added to the `MANIFEST.in` to ensure that distributables (wheels, source distributions) don't include these test files.
2. Custom logging code has been removed from `atlasclient`.  This code simply turned on logging for the underlying modules (requests and urllib) that `atlasclient` is built using. Since `atlasclient` shall be consumed as a library, it will be responsibility of the main application to configure the packages underlying `atlasclient`. Note that this is independent of the client having its own logging functionality - this will be introduced in a separate PR.

### DIFF
--- a/atlasclient/test/test_utils.py
+++ b/atlasclient/test/test_utils.py
@@ -21,25 +21,29 @@ from atlasclient.utils import JSONObject
 
 class TestJSONDecoder(unittest.TestCase):
     def test_simple(self):
-        json_data = {
-            'foo': 1,
-            'bar': 'hello world'}
-        json_obj = JSONObject(json_data)
+        json_data = {'foo': 1, 'bar': 'hello world'}
 
-        for key, value in json_data.items():
-            self.assertEqual(getattr(json_obj, key), json_data[key])
-            self.assertEqual(getattr(json_obj, key), value)
+        # Test both __init__ and from_dict constructors.
+        for mapping_obj in (
+                JSONObject(json_data), JSONObject.from_dict(json_data)):
+            for key, value in mapping_obj.items():
+                self.assertEqual(getattr(mapping_obj, key), json_data[key])
+                self.assertEqual(getattr(mapping_obj, key), value)
+
+    def test_error(self):
+        json_obj = JSONObject.from_dict({})
+        with self.assertRaises(AttributeError):
+            json_obj.a
 
     def test_nested(self):
-        json_data = {
-            'a': {'b': {1: 'foo'}}}
+        json_data = {'a': {'b': {'c': 1}}}
+
+        # Using __init__ only enables dot-access of top level fields.
         json_obj = JSONObject(json_data)
+        with self.assertRaises(AttributeError):
+            self.assertEqual(json_obj.a.b.c, 1)
 
-        def _walk_nested_json_and_assert(jdata, jobj):
-            for key, value in jdata.items():
-                if isinstance(value, dict):
-                    self.assertIsInstance(getattr(jobj, key), JSONObject)
-                    self.assertEqual(getattr(jobj, key), value)
-                    _walk_nested_json_and_assert(value, getattr(jobj, key))
-
-        _walk_nested_json_and_assert(json_data, json_obj)
+        # Using from_dict enables dot-access at all nesting levels.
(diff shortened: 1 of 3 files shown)


atlasclient/utils.py
```diff
@@ -14,29 +14,19 @@
...
-
-from http.client import HTTPConnection
-
 
 class JSONObject(dict):
     """Dictionary with dot-notation read access."""
+    def __coerce_dict(self, value):
+        if isinstance(value, dict):
+            return type(self)(value)
+        return value
```
ShaneHarvey: Does this effectively make nested dicts instead JSONObject immutable? Because we return a JSONObject copy of the actual value.

atlasclient/test/test_utils.py
```diff
@@ -0,0 +1,45 @@
...
+            'bar': 'hello world'}
+        json_obj = JSONObject(json_data)
+
+        for key, value in json_data.items():
+            self.assertEqual(getattr(json_obj, key), json_data[key])
+            self.assertEqual(getattr(json_obj, key), value)
+
+    def test_nested(self):
+        json_data = {
+            'a': {'b': {1: 'foo'}}}
```
ShaneHarvey: Why is `1` a key is this json object? JSON should only have string keys, no?

atlasclient/utils.py
prashantmital: It does indeed. I have now added a classmethod that casts all nested dicts as JSONObjects.

atlasclient/test/test_utils.py
prashantmital: Indeed. Fixed in the new commit.

atlasclient/utils.py
```diff
@@ -14,29 +14,29 @@
...
-
-from http.client import HTTPConnection
+import json
 
 
 class JSONObject(dict):
-    """Dictionary with dot-notation read access."""
+    """Dictionary object with dot-notation read access."""
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
ShaneHarvey: Is this needed?

atlasclient/utils.py
```diff
@@ -14,29 +14,29 @@
...
+        """
+        Create a JSONObject instance from the given dictionary.
+
+        Using this constructor is the recommended way to create JSONObject
+        instances from nested dictionaries as it guarantees the conversion
+        of all nested dicts into JSONObjects. This ensures that
+        dot-notation access works for keys at all nesting levels.
+        """
+        normed_dict = json.loads(json.dumps(raw_dict), object_hook=JSONObject)
+        return cls(normed_dict)
```
ShaneHarvey: I think we can simplify this to just return `normed_dict` because loads does the conversion:
```python
>>> import json
>>> from collections import OrderedDict
>>> m = json.loads(json.dumps({'a':{}}), object_hook=OrderedDict)
>>> type(m), type(m['a'])
(<class 'collections.OrderedDict'>, <class 'collections.OrderedDict'>)
```

atlasclient/utils.py
prashantmital: Nope, not needed. Removed in the next commit.

atlasclient/utils.py
prashantmital: Good point. Done.

atlasclient/utils.py
```diff
@@ -14,29 +14,25 @@
...
 
 
 class JSONObject(dict):
-    """Dictionary with dot-notation read access."""
+    """Dictionary object with dot-notation read access."""
     def __getattr__(self, name):
         if name in self:
             return self[name]
-        raise AttributeError('{} has no property named {}}.'.format(
+        raise AttributeError('{} has no property named {}.'.format(
```
mbroadst: nit: afaict there are no tests around this behavior

prashantmital: added a test for this.

PR #11, opened 2020-03-09
PyMongo Integration

prashantmital:


astrolabe/spec_runner.py
```diff
@@ -177,7 +177,8 @@ def run(self, persist_cluster=False):
         driver_workload = json.dumps(self.spec.driverWorkload)
         worker_subprocess = subprocess.Popen([
             self.config.workload_executor, connection_string,
-            driver_workload], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            driver_workload], preexec_fn=os.setsid,
```
prashantmital: Using `os.setsid` in this manner allows to ensure that all child processes of the script called via Popen inherit the process group ID of the parent. This way, we can send SIGINT to *all* child process when maintenance is done.

temp-tests/retryReads-toggleServerSideJS.yaml
```diff
@@ -0,0 +1,35 @@
+maintenancePlan:
```
prashantmital: These files have been added for convenience. The JS toggle tests are much faster than the cluster resizing ones so its more convenient to iterate on the code using results from just `temp-tests` (the directory containing the two test files)

astrolabe/cli.py
```diff
@@ -174,16 +174,16 @@ def atlas_projects():
     pass
 
 
-@atlas_projects.command('create')
+@atlas_projects.command('ensure')
```
prashantmital: This was changed because more often than not, this is what the user wants to do.

astrolabe/spec_runner.py
```diff
@@ -213,17 +214,30 @@ def run(self, persist_cluster=False):
...
+        LOGGER.debug("stdout:".format(stdout))
+        LOGGER.debug("stderr:".format(stderr))
 
         # Stop the timer
         timer.stop()
 
         # Step-6: compute xunit entry.
         junit_test = junitparser.TestCase(self.id)
         junit_test.time = timer.elapsed
-        if worker_subprocess.returncode != 0:
```
prashantmital: For some reason, on Ubuntu the worker_subprocess was returning a nonzero exit-code. Even after setting a `trap` for the `INT` signal, the return code was nonzero resulting in spurious failures.

prashantmital:
Superseded by #12

PR #12, opened 2020-03-19
PyMongo Integration

prashantmital:


prashantmital:
Passing build: https://evergreen.mongodb.com/version/5e74051a2a60ed78fb798a0f

astrolabe/spec_runner.py
```diff
@@ -213,33 +210,40 @@ def run(self, persist_cluster=False):
         # Step-5: interrupt driver workload and capture streams
         LOGGER.info("Stopping workload executor [PID: {}]".format(
             worker_subprocess.pid))
-        os.kill(worker_subprocess.pid, self.sigint)
-        stdout, stderr = worker_subprocess.communicate()
-        LOGGER.info("Stopped workload executor")
+        os.killpg(worker_subprocess.pid, self.sigint)
```
p-mongo: I know this was not added in this PR, but I believe the standard signal for termination is sigterm, not sigint. The two are often handled similarly but technically they are different signals and especially daemons are not always expecting sigint (since they are not interactive).

prashantmital: Agreed. The reason SIGINT was chosen here instead of SIGTERM is to allow us to be able to differentiate between the termination of the workload executor by the test framework and the termination of the workload executor due to other reasons. SIGINT is always expected to be issued by the user whereas SIGTERM is not. 

Also, I don't believe it is possible to catch SIGTERM and do any special handling for it (which we need to do in order to quit cleanly when the framework requests the workload to stop running and no errors have been encountered). To quote the UNIX wiki: 
> The SIGKILL signal is sent to a process to cause it to terminate immediately (kill). In contrast to SIGTERM and SIGINT, this signal cannot be caught or ignored, and the receiving process cannot perform any clean-up upon receiving this signal.

.evergreen/config.yml
```diff
@@ -0,0 +1,176 @@
...
+          sh .evergreen/install-astrolabe.sh
+
+  "install driver":
+    # Clone driver source code.
+    - command: shell.exec
+      params:
+        script: |
+          ${PREPARE_SHELL_COMMON}
+          git clone --recursive --branch "${DRIVER_REVISION}" "${DRIVER_REPOSITORY}" driver-src
+    # Make the driver's *.sh files executable.
```
mbroadst: this seems like something we shouldn't have to fix in this repository. Can't you make these files executable and commit that to git?

.evergreen/config.yml
```diff
@@ -0,0 +1,176 @@
...
+          for i in $(find .evergreen/atlas-testing -name \*.sh); do
+            chmod +x $i
+          done
+    # Setup driver-specific environment expansions.
+    - command: subprocess.exec
+      params:
+        working_dir: driver-src
+        binary: .evergreen/atlas-testing/write-driver-expansion.sh
+        add_expansions_to_env: true
+    - command: shell.exec
```
mbroadst: I think we can probably get rid of this step too, it looks like debug info?

.evergreen/config.yml
```diff
@@ -0,0 +1,176 @@
...
+        script: |
+          ${PREPARE_SHELL_COMMON}
+          for i in $(find .evergreen/atlas-testing -name \*.sh); do
+            chmod +x $i
+          done
+    # Setup driver-specific environment expansions.
+    - command: subprocess.exec
+      params:
+        working_dir: driver-src
+        binary: .evergreen/atlas-testing/write-driver-expansion.sh
```
mbroadst: Is this necessary? I'm trying to think with my "how do we get this done for 12 drivers" hat here, could we fold both of these scripts into something like `install-dependencies.sh` which _always_ writes a `driver-expansions.yaml` file, even if its empty

astrolabe/spec_runner.py
```diff
@@ -213,33 +210,40 @@ def run(self, persist_cluster=False):
...
 
         # Step-6: compute xunit entry.
         junit_test = junitparser.TestCase(self.id)
         junit_test.time = timer.elapsed
-        if worker_subprocess.returncode != 0:
+
+        try:
+            err_info = json.loads(stderr)
+        except json.JSONDecodeError:
+            err_info = {'numErrors': '-1', 'numFailures': '-1'}
```
ShaneHarvey: Can we change these to be numbers (`'-1'` to `-1`) so that we don't need to call `int` later?

astrolabe/spec_runner.py
```diff
@@ -213,33 +210,40 @@ def run(self, persist_cluster=False):
...
             Number of errors: {numErrors}
-            Number of failures: {numFailures}    
-            """
-            try:
-                err_info = json.loads(stderr)
-                junit_test.result = junitparser.Failure(
-                    errmsg.format(**err_info))
-            except json.JSONDecodeError:
-                junit_test.result = junitparser.Error(encode_cdata(stderr))
+            Number of failures: {numFailures}
```
ShaneHarvey: This error message uses multi-line string so it includes all the whitespace. Do we want it to be indented this much?

astrolabe/spec_runner.py
```diff
@@ -52,49 +52,45 @@ def __init__(self, *, client, test_name, cluster_name, specification,
...
+            prefix, suffix = cluster.srvAddress.split("//")
+            uri_options = self.spec.maintenancePlan.uriOptions.copy()
+
+            # Boolean options must be converted to lowercase strings.
+            for key, value in uri_options.items():
+                if isinstance(value, bool):
+                    uri_options[key] = str(value).lower()
+
+            connection_string = (prefix + "//" + self.config.database_username
+                                 + ":" + self.config.database_password + "@"
```
ShaneHarvey: Don't we need to url encode the username and password here?

.evergreen/config.yml
```diff
@@ -0,0 +1,176 @@
...
+    - command: expansions.update
+      params:
+        file: driver-src/driver-expansion.yml
+    # Run driver-supplied script to install the driver.
+    - command: shell.exec
+      params:
+        working_dir: driver-src
+        script: |
+          ${PREPARE_SHELL_COMMON}
+          ${PREPARE_SHELL_DRIVER}
```
ShaneHarvey: Where is `PREPARE_SHELL_DRIVER` defined?

.evergreen/config.yml
```diff
@@ -0,0 +1,176 @@
...
+  # The 'driver' axis specifies the driver to be tested (including driver version).
+  # Drivers MUST specify the DRIVER_REPOSITORY and DRIVER_REVISION variables.
+  - id: driver
+    display_name: driver
+    values:
+      - id: pymongo-master
+        display_name: "PyMongo (master)"
+        variables:
+          DRIVER_REPOSITORY: "https://github.com/prashantmital/mongo-python-driver.git"
+          DRIVER_REVISION: "PYTHON-1743/test-atlas-planned-maintenance"
```
ShaneHarvey: I thought the plan was to add the driver specific test scripts to this repo. Why did we decide to add the scripts to the driver repo instead? This makes it more difficult to follow the flow because it jumps back and forth between multiple files in two repos, eg PYTHON_BINARY is defined in 'runtime' axis and used in `mongo-python-driver/.evergreen/atlas-testing/install-driver.sh`.

.evergreen/config.yml
```diff
@@ -0,0 +1,176 @@
...
+        display_name: CPython-3.8
+        variables:
+          PYTHON_BINARY: "/opt/python/3.8/bin/python3"
+
+buildvariants:
+- matrix_name: "tests-python"
+  matrix_spec:
+    driver: ["pymongo-master"]
+    platform: ["ubuntu-16.04"]
+    runtime: ["python36", "python37", "python38"]
```
ShaneHarvey: We should add 2.7 too. If there is some complexity adding 2.7 we should open a follow up ticket for it.

.evergreen/config.yml
prashantmital: Yes, we can require that to be done by drivers. I will add a note to the docs/integration guide.

.evergreen/config.yml
prashantmital: We _could_ get rid of it but it is usually quite helpful during debugging to not have to edit the config and re-run the patch just to figure out what some expansions evaluated to.

.evergreen/config.yml
prashantmital: This is defined in driver-expansion.yml which is written when the driver-supplied write-driver-expansions.sh script is executed.

.evergreen/config.yml
prashantmital: Good idea - the install-driver and write-driver-expansions scripts should be able to be combined without any issues. I will double-check before committing to the change though.

.evergreen/config.yml
prashantmital: Yes it definitely makes it harder to follow the flow, but this was done to minimize code churn in the driver-atlas-testing repo when more drivers start integrating this testing into their CI workflows. I think there were other, good, reasons we made this change too (originally the driver-specific scripts lived in driver-specific folders in this repository) but I cannot recall at the moment. @mbroadst do you recall?

.evergreen/config.yml
prashantmital: Will do. I don't think there is any complexity associated with adding that but it wasn't added here because the framework itself is Py3-only and I got into the bad habit of installing PyMongo into the same env as astrolabe during testing.

astrolabe/spec_runner.py
prashantmital: I think you are right. I didn't pay any special attention to it since we will mostly just end up using the default userpass. I will change this in the next commit.

astrolabe/spec_runner.py
prashantmital: What about the case when we parse output returned by the workload executor? I was under the impression that the numbers in the JSON returned by that script would also end up getting parsed as string so I needed to perform this type-conversion regardless.

astrolabe/spec_runner.py
prashantmital: Good point, I will remove the whitespace.

.evergreen/config.yml
mbroadst: @ShaneHarvey I was worried that we would have a situation like we do with libmongocrypt here, where 12 different drivers were committing with different styles to a common repository. Rather, this project has inverted that design to expect certain "hooks" to be present in a given driver's repository. The hope was also that drivers would be able to reuse a lot of scripts they already have (e.g. node has an `install-dependencies.sh` script that already downloads a version of node and installs it globally), so the hook could just call that. 

At the end of the day I think the most complicated part is the case you bring up, but I'm not sure solving for that particular issue provides the most gain in terms of ease of implementation for all drivers. My expectation is that the variables you use in the axes here are things you already depend upon in your existing evergreen config.

ShaneHarvey: > where 12 different drivers were committing with different styles to a common repository. 

Hm, I see that as a non-issue especially because the driver tests are going to be released/published anywhere.

Another benefit of having it all live in this repo is that it becomes trivial to test different branches/versions of the driver in question (ie we could test pymongo 3.10, 3.11, master, etc...). With the current approach we would need to commit and maintain the test scripts in all branches we want to test.

.evergreen/config.yml
ShaneHarvey: Yeah, doesn't astrolabe itself use pymongo? So the versions could clash?

astrolabe/spec_runner.py
ShaneHarvey: `json.loads` does the number parsing:
```
>>> import json
>>> json.loads('{"num": -1}')
{'num': -1}
```

.evergreen/config.yml
prashantmital: In the way the evergreen job is setup, this is a non-issue as I am using different virtualenvs for astrolabe and pymongo. When testing locally, that could certainly happen but it wasn't a major concern in the prototyping phase.

.evergreen/config.yml
prashantmital: I am in the process of changing this so that driver-specific scripts live in this repo again. I agree with your rationale @ShaneHarvey - this is a much better pattern.

ShaneHarvey: Cool! Just one final note. Matt mentioned that he might prefer these tests to live in the node driver rather than this project for a couple reasons I won't list. But can't we have it both ways? PyMongo's tests can live in this repo and Node live in another repo.

mbroadst: @ShaneHarvey we're just going to move everything over to this repo, I'd prefer to have a single solution for all drivers at this point rather than get fancy.

.evergreen/config.yml
prashantmital: Actually, on a closer look, it will not be possible to combine these two files into a single file. The reason being that the install-driver.sh step might _require_ some of the expansions created by the write-driver-expansion.yml step. Unfortunately, this means that we must call the expansions.update evergreen command against the YML file written by the script **before** invoking install-driver.sh.

prashantmital: Scratch that, I managed to rework things so that the driver can be installed using a single script.

.evergreen/pymongo/workload-executor.py
```diff
@@ -0,0 +1,95 @@
+import copy
```
mbroadst, line 1: slight preference for `.evergreen/<language>` rather than custom naming (similar to [libmongocrypt](https://github.com/mongodb/libmongocrypt/tree/master/bindings)). We could even do `.evergreen/<language>/<named project>` for cases where you have multiple flavors (e.g. `motor`, `pymongo`)

.evergreen/config.yml
ShaneHarvey: I think you forgot to add Python 2.7 here.

.evergreen/pymongo/workload-executor.py
```diff
@@ -0,0 +1,95 @@
...
+    result = cmd(**dict(arguments))
+
+    if expected_result is not None:
+        if isinstance(result, Cursor) or isinstance(result, CommandCursor):
+            result = list(result)
+        assert result == expected_result
+
+
+def workload_runner(srv_address, workload_spec):
+    # Do not modify connection string and do not add option of your own.
```
ShaneHarvey, line 49: Nit/typo: "option of your own." -> "any extra options".

mbroadst:
@prashantmital what's holding this up at this point?

.evergreen/pymongo/workload-executor.py
prashantmital: Done!

.evergreen/config.yml
prashantmital: Done.

PR #17, opened 2020-04-06
Add Integration Guide

### LINKED ISSUE #15: Add integration guide
At a minimum, documentation that delineates how a new driver may use astrolabe to implement atlas planned maintenance testing needs to added.

prashantmital:
Closes #15


prashantmital:
I can set things up so that the docs will be automatically built whenever `master` is updated and then hosted on GitHub pages. I already did this in my fork to show you what it would look like: https://prashantmital.github.io/drivers-atlas-testing/

Let me know if I should go ahead and do the same here.

docs/source/integration-guide.rst
```diff
@@ -0,0 +1,189 @@
...
+----------------------------------------------
+Step 2: Implementing a Driver Installer Script
+----------------------------------------------
+
+Drivers must implement this standalone script to perform all setup/installation-related tasks for their driver.
+The installer script MUST be saved in the ``DRIVER_DIRNAME`` directory under the name ``install-driver.sh``.
+The executable permission bit MUST be set for the install script file before it is committed to git.
+
+This script can be used to perform any number of arbitrary tasks related to setting up the environment for
+the workload executor to be executed within. It MUST NOT however, clone the driver source repository.
```
mbroadst, line 103: Please explain why this "MUST NOT" is included.

docs/source/integration-guide.rst
```diff
@@ -0,0 +1,189 @@
...
+
+.. note::
+
+   All Evergreen expansions for a given build variant are available to the installer script
```
mbroadst, line 107: didn't we decide that this file itself could output expansion updates? I don't think I see that discussed here.

docs/source/integration-guide.rst
```diff
@@ -0,0 +1,189 @@
...
+   at runtime as environment variables. Scripts can make use of any of these environment variables but must
+   ensure that they are written in a way that prevents accidentally leaking Atlas API credentials and
+   MongoDB user credentials.
+
+
+.. note::
+
+   Even if a driver does not require an installer script, it MUST create an empty/no-op script of the required
```
mbroadst, line 115: I think we can probably work around this in the evergreen configuration? Just test for the existence of the file and only run it if it is present?

docs/source/index.rst
```diff
@@ -0,0 +1,23 @@
+Welcome to astrolabe's documentation!
```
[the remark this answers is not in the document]
prashantmital: No, we don't have to publish elsewhere but I used RST because we use it for our specifications. At this point its more effort to switch it back so I'd be reluctant to do so unless there's some major problem with RST.

docs/source/integration-guide.rst
```diff
@@ -0,0 +1,189 @@
...
+#. Implement a Driver Installation script for their driver project.
+#. Update the Evergreen configuration of the ``drivers-atlas-testing`` project.
+
+
+We will walk through each of these steps in detail in the following sections.
+
+Setting up
+----------
+
+#. Start by cloning the ``drivers-atlas-testing`` repository and creating a new branch.
```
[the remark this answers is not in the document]
prashantmital: Actually, I forgot a step here (to fork the repo first, then clone, then branch). In that light, the code-samples wont exactly be copy-pastable since we have no way to pre-empt the GH usernames of folks who will be using this, but I have added samples with placeholders nonetheless.

docs/source/integration-guide.rst
```diff
@@ -0,0 +1,189 @@
...
+* MUST log all output from the driver during the test run, including any tracebacks, and informational messages to STDOUT.
+* MUST NOT override any of the URI options specified in the incoming connectionString.
+* MUST NOT augment the incoming connectionString with any additional URI options.
+
+Finally, the workload executor MUST be saved in the ``DRIVER_DIRNAME`` directory under the name ``workload-executor.sh``.
+The executable permission bit MUST be set for the workload executor file before it is committed to git.
+
+.. note::
+
+   All Evergreen expansions for a given build variant are available to the workload executor script
```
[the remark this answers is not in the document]
prashantmital: I have added a link to Evergreen's documentation about expansions and added some more information in a note in the 'Evergreen configuration' section. It doesn't make much sense to document what expansions/variables are available as none of the driver-implemented components are supposed to rely on any of them. This statement is included here primarily to indicate that any variables that the driver engineer has defined in their EVG build matrix would be available to these scripts as environment variables.

docs/source/integration-guide.rst
```diff
@@ -0,0 +1,189 @@
...
+   It is recommended that drivers use the name of their language to name their folder.
+   Languages that have multiple drivers should use subdirectories named after the driver (e.g. for Python,
+   we should create ``.evergreen/python/pymongo`` and ``.evergreen/python/motor``). The path to the newly created
+   folder(s) will henceforth be referred to in this guide as ``DRIVER_DIRNAME``.
+
+----------------------------------------
+Step 1: Implementing a Workload Executor
+----------------------------------------
+
+Drivers must implement this standalone script in accordance with the specification, see :ref:`workload-executor-specification`.
```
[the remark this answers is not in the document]
prashantmital: The referenced section describes the public API of the workload executor and also provides a pseudocode example (its basically the workload executor section of the specification)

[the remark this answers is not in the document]
prashantmital: The implementation guide doesn't inform the user of the API required for the workload executor since that would duplicate information and necessitate its maintenance in two places.

docs/source/integration-guide.rst
prashantmital: We no longer need to write a file to do expansion updates. We use `subprocess.exec`(https://github.com/evergreen-ci/evergreen/wiki/Project-Commands#subprocessexec) with the `add_expansions_to_env` option so we don't need to mess around with the clunky expansions.yml files.

docs/source/integration-guide.rst
prashantmital: Good point. Done.

docs/source/index.rst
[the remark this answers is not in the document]
mbroadst: Well my problem with RST is that we now need to publish this somewhere, right? Or is your intention that people just read this in the repo? My thought with converting to Markdown for the README.md is that it's implicitly published on the main page of the repo

docs/source/integration-guide.rst
mbroadst: so you're saying that any environment variable I set in my `install-driver.sh` script is automatically added to expansions that I can use in the next phase? If that's not the case, how does someone implementing this expose environment variables to use in the next phase if needed?

docs/source/integration-guide.rst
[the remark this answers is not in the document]
mbroadst: Ok, thanks!

docs/source/integration-guide.rst
prashantmital: Actually, you need to set the environment variables when setting up your driver and/or runtime in the evergreen configuration. I will make a note of it here so that it is clear. E.g. `PYTHON_BINARY` and `PYMONGO_VIRTUALENV_NAME` are variables that are need by pymongo's installer and workload executor scripts and you can see where we define them in `.evergreen/config.yml`.

mbroadst: Could it be the case that some driver needs to dynamically set a variable? What if part of `install-driver.sh` for the Rust driver uses `rustup` to install a custom version of Rust, and then needs to refer to that location in the `workload-executor.sh` script?

prashantmital: Yes, I suppose this could be needed. That being said, usually the tool like `rustup` will provide a CLI-based method to retrieve the path of what was installed, OR it would let you invoke your executable with what was downloaded. In the case of `rustup` for instance, if you want to run the test with rust nightly, you can do `rustup run nightly rustc foo.rs`.

To put it another way, I am not entirely convinced that we should add the scaffolding for expansion files unless absolutely necessary and at this point, I am not sure that we know it is necessary.

docs/source/index.rst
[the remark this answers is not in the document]
prashantmital: Yes, I will publish it using GH pages.

PR #50, opened 2020-04-24
Use 'Project' instead of 'Group' everywhere

### LINKED ISSUE #25: Refer to Atlas Projects as projects everywhere
Currently, we use a mix of 'group' and 'project' to refer to the same thing. "Project" is more user-friendly and self-explanatory and we should switch to using that entirely.

prashantmital:
Closes #25


prashantmital:
Passing patch build: https://evergreen.mongodb.com/version/5ea283430305b9243b48a72f

PR #51, opened 2020-04-24
Report workload statistics using a file instead of using stderr

### LINKED ISSUE #19: Workload executors should report JSON output using a file
Currently, we require workload executors to report statistics of the run back to the test orchestrator using JSON output that is sent to STDERR. This method of doing things breaks down however in the event that the workload produces any output that is sent to STDERR during the test run. Some languages don't provide any mechanism to redirect such output to STDOUT.

To accommodate all languages, we can instead require this JSON output to be written as ASCII text by the workload executor to a file with a fixed, pre-determined name in the current working directory where the workload executor is invoked. The orchestrator can parse this file for the desired information once the workload executor has been terminated. If the file is not found, the orchestrator can assume that something went wrong and therefore mark the run as a failure.

prashantmital:
Closes #19

### DIFF
--- a/.evergreen/python/pymongo/workload-executor.py
+++ b/.evergreen/python/pymongo/workload-executor.py
@@ -70,21 +70,21 @@ def run_operation(objects, prepared_operation):
         assert result == expected_result
 
 
-def connect(connection_string):
+def connect(mongodb_uri):
     if WIN32:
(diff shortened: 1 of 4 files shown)


prashantmital:
Passing patch build: https://evergreen.mongodb.com/version/5ea28f180305b9243b48bd03

PR #57, opened 2020-04-30
Ensure workload executor sentinel file results.json is cleaned up

prashantmital:


astrolabe/utils.py
```diff
@@ -175,7 +182,7 @@ def terminate(self):
         outs, errs = self.workload_subprocess.communicate(timeout=10)
 
         try:
-            with open('results.json', 'r') as fp:
+            with open(self.sentinel, 'r') as fp:
                 stats = json.load(fp)
         except (FileNotFoundError, json.JSONDecodeError):
             stats = {'numErrors': -1, 'numFailures': -1}
```
mbroadst, line 188: I would imagine we want to warn very loudly if a workload executor fails to produce a `results.json`

Issue #48, opened 2020-04-24
Workaround Atlas limitation on maximum number of Clusters under a Project


prashantmital:
Atlas limits each Project to 25 clusters. This means we are limited to 25 concurrent builds at any time in the drivers-atlas-testing evergreen project. In the absence of this restriction, we often end up with too many simultaneously executing jobs that end up failing with the `CLUSTER_NOT_FOUND` error code (see #47 for details). 

We need to figure out a workaround for this limitation. CC: @mbroadst

prashantmital:
One option could be to let each driver use their own Atlas Project to run their tests under but that would still restrict us to 25 concurrent builds per language. Enforcing this limit on a per-language basis might be tougher than enforcing it across all languages.

prashantmital:
Using Evergreen's `task_group` function () with the `max_hosts` option set appropriately might be another possible solution to this problem. 

UPDATE: this might not actually be a solution since each build variant gets its own copy of the `task_group` which, in turn, distributes tasks across a maximum of `max_hosts` hosts. That means that if we have `m` buildvariants and `max_hosts=k`, we could potentially have `m*k` tasks running concurrently at any given time. Thus we'd need to ensure `m*k <= 25` which isn't an easy condition to meet given that the number of build variants (`m`) driver languages would add to this project is potentially unbounded.

prashantmital:
Further confirmation that we are running into this limitation (and that the failures are not caused by problems in the Python driver and/or bugs in `astrolabe`): [all the tests are green](https://evergreen.mongodb.com/version/drivers_atlas_testing_563e7ce4e286e1555a69dfb315d1edb720812fbd) if we restart the failing tests after a run has completed. This is because we have around 40 tests. Around 15 failed during the initial run (maximum of 25 clusters can be created at a time). These 15 when restarted only request 15 clusters on an Atlas project that has no other running clusters and are therefore able to go through to completion. 

Patch build: https://evergreen.mongodb.com/waterfall/drivers-atlas-testing

prashantmital:
Note that #65 didn't completely resolve this issue. While it smooths over things for time being, once more entries in the build matrix start using other distros than the custom one added in #65, we will start hitting this issue again. I have opened a [feature request](https://jira.mongodb.org/browse/EVG-7951) to see if the Evergreen team can come up with a more permanent fix.

PR #58, opened 2020-04-30
Reduce number of test cases for PyMongo

prashantmital:
There doesn't seem to be an imminent need for all these test cases for Python. Testing CPython 2.7 and 3.8 is sufficient on Linux as is testing 3.7 on Windows. Also, we can defer adding testing for released versions (in this case 3.10.x) to a later time when the need presents itself - testing master is sufficient for the time being.


prashantmital:
Note that this helps bring the number of cases under 25 which should allow Evergreen to be green for the time being (see #48).

prashantmital:
Passing patch: https://evergreen.mongodb.com/version/5eaa18e832f417570a997fb3

PR #59, opened 2020-04-30
Use non-dotted directories for workload executors

### LINKED ISSUE #53: Use non-dotted directories for workload executors
Currently workload executors are located in subdirectories of `.evergreen`. In Unix the leading dot makes a file hidden, which means:

- Commands like `ls` do not show these files/directories by default
- Shell completion by default does not show them

Also, the leading dot is awkward to type. I use Dvorak layout but even on qwerty "a" is on the home row and "." is one row down. On Dvorak . and e are adjacent (same finger) and are arranged vertically.

Is it possible to have the workload executors be under a path that does not contain  dotted directories please?

prashantmital:
Closes #53 

Instead of using the `workloads` name for the new directory as suggested in the ticket, I used the name `integrations` because the folder contains everything related to a driver's integration which is more than simply the workload.


prashantmital:
Passing patch: https://evergreen.mongodb.com/version/5eaa328461837d5baf37b9ce

prashantmital:
@mbroadst if you approve, I will move the evergreen config file to `evergreen.yml` at the root of the repo instead of where it is currently located. It no longer makes sense to have a `.evergreen` folder as everything other than the configuration file will live in `integrations`.

mbroadst:
@prashantmital okay

prashantmital:
Not moving the configuration file in light of our conversation earlier today @mbroadst (the `.evergreen` might be needed to hold scripts related to the workload executor validator task).

Issue #56, opened 2020-04-27
Investigate why Atlas Group retrieval fails for some groups


prashantmital:

prashantmital:
This is an Atlas API issue. See https://jira.mongodb.org/browse/HELP-15367 for details.

prashantmital:
Closing as this will not be solved by ``astrolabe`` and will need to be addressed upstream.

PR #60, opened 2020-05-01
Make astrolabe resilient to CLOUDP-61574

prashantmital:
Related to #56 

Enables astrolabe to parse NOT_ATLAS_GROUP error response for the ID of the Atlas Project being requested. The main idea here is that we need the project ID in order to perform an operation on the project (e.g. populating its IP whitelist) in order for Atlas to successfully populate the project metadata so that we may use the project as normal.

Just opening this for visibility at the moment, since this 'workaround' isn't actually sufficient to resolve the issues we face when using the Atlas API. Follow https://jira.mongodb.org/browse/HELP-15367 for details.

### DIFF
--- a/astrolabe/cli.py
+++ b/astrolabe/cli.py
@@ -201,7 +201,10 @@ def list_projects(ctx):
 @click.pass_context
 def get_one_project_by_name(ctx, project_name):
     """Get one Atlas Project."""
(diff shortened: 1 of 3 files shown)


prashantmital:
This was a specific issue related to one of the account we use to orchestrate the clusters used for the drivers-atlas-testing tests. It has since been resolved by Atlas.

PR #61, opened 2020-05-01
Add command to validate workload executors

### LINKED ISSUE #6: Support quick-finishing when workload-executor throws an error
When the workload executor is incorrectly implemented or error-ing due to any other reason, the test runner should cut short test execution, report the error raised by the executor via JUnit, and then move onto the next test.

Currently, the test runner doesn't check if the workload executor has errored and keeps waiting for cluster maintenance to complete even if the driver workload is not being run.

### LINKED ISSUE #35: Mechanism for validating workload executors
Add a command to astrolabe that takes a `WORKLOAD_EXECUTOR` and a connection string and runs a series of workloads that are known to fail as a smoketest for checking whether the Workload Executor works properly and returns the correct JSON output. 

This would be really helpful to have as it can be quite hard to figure out whether the workload executor is working correctly especially when there are no failures. Since the workload executor runs in a background process, on the off chance that it fails silently (due to an implementation bug usually), we could end up getting spurious 'passing'/'green' builds.

The validator would run a bunch of invalid workloads to stress-test the workload executor and to check that it raises the expected errors (e.g. run a non-existent command `N` times and and check that the workload executor returns `numErrors=N`). Additionally, we can also run queries that succeed and then have `astrolabe` check that we see the expected changes on the server (e.g. issue a write via the workload executor and then check using `astrolabe` that the expected document shows up on the server).

prashantmital:
Closes #35 
Closes #6

### DIFF
--- a/docs/source/spec-workload-executor.rst
+++ b/docs/source/spec-workload-executor.rst
@@ -59,6 +59,7 @@ After accepting the inputs, the workload executor:
    * MUST keep count of the number of operation errors (``numErrors``) that are encountered while running
      operations. An operation error is when running an operation unexpectedly raises an error. Workload executors
      implementations should try to be as resilient as possible to these kinds of operation errors.
+   * MUST keep count of the number of operations that are run successfully (``numSuccesses``).
 
 #. MUST set a signal handler for handling the termination signal that is sent by ``astrolabe``. The termination signal
    is used by ``astrolabe`` to communicate to the workload executor that it should stop running operations. Upon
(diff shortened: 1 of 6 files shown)


docs/source/spec-workload-executor.rst
```diff
@@ -71,15 +72,16 @@ After accepting the inputs, the workload executor:
 
      * ``numErrors``: the number of operation errors that were encountered during the test.
      * ``numFailures``: the number of operation failures that were encountered during the test.
+     * ``numSuccessfulOperations``: the number of operations executed successfully during the test.
```
mbroadst, line 75: nit: `numSuccessfulOperations` lacks symmetry with the other two properties here, why not just `numSuccesses`. Alternatively we could be more explicit with all three: `errorCount`, `failureCount`, `successCount`

prashantmital: Good point. I have made this change in the latest commit.

prashantmital:
To avoid making this PR overly large, I will tackle documentation and the evergreen task to validate workload executors in #66 and #67

prashantmital:
Passing patch: https://evergreen.mongodb.com/task/drivers_atlas_testing_tests_python__driver~pymongo_master_platform~ubuntu_18.04_runtime~python27_retryWrites_toggleServerSideJS_patch_6c51b510c8783b7a3999f2ed651b73b73c57a10e_5eb19b533627e04b03e891aa_20_05_05_17_00_05

PR #63, opened 2020-05-01
Stop redirecting standard streams of the workload executor subprocess

### LINKED ISSUE #9: Make it safe for workload executors to output large amounts of text
Currently, when running the workload executor, we PIPE the output from its stderr and stdout streams. To avoid blocking the child process when the pipe fills up, we need to continuously read the worker_subprocess output (via worker_subprocess.stdout.read(..)).

Thanks to @ShaneHarvey for the context:

> See the note in subprocess.wait:
> 
> This will deadlock when using stdout=PIPE or stderr=PIPE and the child process generates enough output to a pipe such that it blocks waiting for the OS pipe buffer to accept more data. Use Popen.communicate() when using pipes to avoid that.
> 
> https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait
> 
> Just to be clear: there's no risk of deadlock in the current approach. The risk is that the child process can fill up the pipe with output and then it will block until we call worker_subprocess.communicate() down below.

A more elegant solution will be to redirect the worker_subprocess output to a file. Then we can open and read the file after the process exits.

prashantmital:
Closes #9 

See https://github.com/mongodb-labs/drivers-atlas-testing/issues/9#issuecomment-622610173 for rationale.

### DIFF
--- a/astrolabe/exceptions.py
+++ b/astrolabe/exceptions.py
@@ -22,3 +22,7 @@ class AstrolabeTestCaseError(AstrolabeBaseError):
 
 class PollingTimeoutError(AstrolabeBaseError):
     pass
+
+
+class WorkloadExecutorError(AstrolabeBaseError):
+    pass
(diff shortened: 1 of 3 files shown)


prashantmital:
Passing patch: https://evergreen.mongodb.com/version/5eacb366e3c3315f26d77e0e

PR #65, opened 2020-05-05
Migrate linux builds to ubuntu1804-drivers-atlas-testing

### LINKED ISSUE #48: Workaround Atlas limitation on maximum number of Clusters under a Project

### LINKED ISSUE #62: Migrate to  ubuntu1804-drivers-atlas-testing
This distro is limited to 25 instances so it helps us workaround #48.

We must also document that drivers need to use this distro to write their Linux-based test cases.

prashantmital:
Closes #62 
Closes #48 (kind of - it doesn't actually solve this problem in a foolproof manner because if we add more platforms, we could end up running more than N concurrent builds where N is the maximum number of instances of `ubuntu1804-drivers-atlas-testing`).

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -24,6 +24,12 @@ functions:
     - command: git.get_project
       params:
         directory: astrolabe-src
+    # Make sure virtualenv is installed.
+    - command: subprocess.exec
+      params:
+        working_dir: astrolabe-src
+        command: |
+          ${PYTHON3_BINARY} -m pip install virtualenv
     # Create virtualenv using a CPython 3.5+ binary.
     - command: subprocess.exec
       params:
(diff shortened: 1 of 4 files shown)


prashantmital:
Passing patch: https://evergreen.mongodb.com/version/5eb0d279a4cf47061ea31e77

prashantmital:
Passing patch: https://evergreen.mongodb.com/version/5eb0d279a4cf47061ea31e77

PR #68, opened 2020-05-06
Enable running workload executor validation as an evergreen patch-build

### LINKED ISSUE #66: Workload executor validation evergreen task
Need to add a patch-only task to run the workload executor validation in evergreen. This will probably be easier to use for driver engineers as they won't have to manually set up environment variables and download/install the respective drivers (instead we will piggyback off of `install-driver.sh` and the usual scaffolding that is used in production builds).

### LINKED ISSUE #67: Add documentation for workload executor validation
Documentation on how to use this feature locally and in evergreen (see #66) is required.

prashantmital:
Closes #66 
Closes #67

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -82,6 +82,22 @@ functions:
         command: |
           astrolabevenv/${PYTHON_BIN_DIR}/astrolabe spec-tests run-one tests/${TEST_NAME}.yaml -e integrations/${DRIVER_DIRNAME}/workload-executor
 
+  "validate executor":
+    # Run a MongoDB instance locally.
+    - command: shell.exec
+      params:
+        working_dir: astrolabe-src
+        script: |
+          sh .evergreen/run-mongodb.sh
+    # Validate the workload executor against the local MongoDB instance.
+    - command: subprocess.exec
+      type: test
+      params:
+        working_dir: astrolabe-src
+        add_expansions_to_env: true
+        command: |
+          astrolabevenv/${PYTHON_BIN_DIR}/astrolabe spec-tests validate-workload-executor -e integrations/${DRIVER_DIRNAME}/workload-executor --connection-string "mongodb://localhost:27017/"
+
   "delete test cluster":
     # Delete the cluster that was used to run the test.
     - command: subprocess.exec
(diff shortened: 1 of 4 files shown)


prashantmital:
See it in action: https://evergreen.mongodb.com/version/5eb269162a60ed7de813af11

.evergreen/config.yml
```diff
@@ -82,11 +82,60 @@ functions:
         command: |
           astrolabevenv/${PYTHON_BIN_DIR}/astrolabe spec-tests run-one tests/${TEST_NAME}.yaml -e integrations/${DRIVER_DIRNAME}/workload-executor
 
+  "validate executor":
+    # Run a MongoDB instance locally.
+    - command: shell.exec
+      params:
+        working_dir: astrolabe-src
+        script: |
+          set -o xtrace
```
mbroadst, line 91: This section looks like a copy of `run-mongodb.sh`, is that a mistake?

.evergreen/run-mongodb.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/sh
+set -o xtrace
+
+# User configurable-options
+export MONGODB_VERSION="4.2"
+export TOPOLOGY="replica_set"
```
mbroadst, line 6: why run against a replicaset? I don't think we're testing anything complicated, I would just start up a single server

.evergreen/config.yml
```diff
@@ -82,11 +82,60 @@ functions:
...
+        add_expansions_to_env: true
+        command: |
+          astrolabevenv/${PYTHON_BIN_DIR}/astrolabe spec-tests validate-workload-executor -e integrations/${DRIVER_DIRNAME}/workload-executor --connection-string "mongodb://localhost:27017/?replicaSet=repl0"
+
   "delete test cluster":
     # Delete the cluster that was used to run the test.
     - command: subprocess.exec
       params:
         working_dir: astrolabe-src
+        continue_on_err: true       # Because we may not be running an actual spec test.
```
mbroadst, line 138: this feels like a kludge to me, we shouldn't be cleaning up a cluster if one was never created (for the validation task)

.evergreen/config.yml
prashantmital: Moved everything to `run-mongodb.sh`.

.evergreen/run-mongodb.sh
prashantmital: Changed to standalone.

prashantmital:
Failures are due to #71, I will address them separately as it is an astrolabe internal change.

.evergreen/config.yml
prashantmital: Removed the `continue_on_err`. Its not required as failures in `post` are ignored anyway.

PR #75, opened 2020-05-20
node integration

emadum:
Adds workload-executor integration for Node, and runtimes for the 3 most recent LTS versions of node (Carbon/v8, Dubnium/v10, Erbium/v12).

[evergreen patch](https://evergreen.mongodb.com/version/5ec6868dc9ec440628ee9129)

### DIFF
--- a/integrations/node/workload-executor.js
+++ b/integrations/node/workload-executor.js
@@ -5,38 +5,48 @@ const fs = require('fs');
 const assert = require('assert');
 const omit = require('lodash.omit');
 
+const NOOP_DELAY = 1000;
+
 let interrupted = false;
 process.on('SIGINT', () => (interrupted = true));
 
-const executors = {
-  find: (collection, args) =>
-    collection.find(args.filter || {}, omit(args, ['filter'])).toArray(),
-
-  insertOne: (collection, args) =>
-    collection.insertOne(args.document, { forceServerObjectId: true }),
-
-  updateOne: (collection, args) =>
-    collection.updateOne(args.filter, args.update)
-};
+const operations = new Map();
+operations.set(
+  'find',
+  (collection, args) => collection.find(args.filter || {}, omit(args, ['filter'])).toArray()
+);
+operations.set(
+  'insertOne',
+  (collection, args) => collection.insertOne(args.document /*, { forceServerObjectId: true }*/)
+);
+operations.set(
+  'updateOne',
+  (collection, args) => collection.updateOne(args.filter, args.update)
+);
 
 async function main(uri, spec) {
   const runOperations = makeRunOperations(spec);
 
   const client = new MongoClient(uri);
-  await client.connect();
+  try {
+    await client.connect();
 
-  const results = { numSuccesses: 0, numFailures: 0, numErrors: 0 };
+    const results = { numSuccesses: 0, numFailures: 0, numErrors: 0 };
 
-  while (!interrupted) {
-    try {
-      await runOperations(client, results);
-    } catch (err) {
-      ++results.numErrors;
+    while (!interrupted) {
+      try {
+        await runOperations(client, results);
+      } catch (err) {
+        ++results.numErrors;
+      }
     }
-  }
 
-  fs.writeFileSync('results.json', JSON.stringify(results));
-  await client.close();
+    fs.writeFileSync('results.json', JSON.stringify(results));
+  } catch (error) {
+    console.error(error);
+  } finally {
+    await client.close();
+  }
 
   process.exit(0);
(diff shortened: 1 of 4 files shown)


integrations/node/workload-executor.js
```diff
@@ -0,0 +1,100 @@
...
+
+let interrupted = false;
+process.on('SIGINT', () => (interrupted = true));
+
+const executors = {
+  find: (collection, args) =>
+    collection.find(args.filter || {}, omit(args, ['filter'])).toArray(),
+
+  insertOne: (collection, args) =>
+    collection.insertOne(args.document, { forceServerObjectId: true }),
```
emadum, line 16: Note - I get a bunch of duplicate key errors without `forceServerObjectId: true`.

mbroadst: can you paste the errors here? That's very strange, I wouldn't expect that to happen.

integrations/node/workload-executor.js
```diff
@@ -0,0 +1,100 @@
...
+  await client.close();
+
+  process.exit(0);
+}
+
+function makeRunOperations(spec) {
+  const operations = spec.operations;
+  return function(client, results) {
+    if (!Array.isArray(operations) || operations.length === 0) {
+      return new Promise(resolve => setTimeout(resolve, 1000));
```
mbroadst, line 48: can we use a more descriptive constant for this magic 1000 number?

integrations/node/workload-executor.js
```diff
@@ -0,0 +1,100 @@
...
+
+const MongoClient = require('../../node-mongodb-native').MongoClient;
+const fs = require('fs');
+const assert = require('assert');
+const omit = require('lodash.omit');
+
+let interrupted = false;
+process.on('SIGINT', () => (interrupted = true));
+
+const executors = {
```
mbroadst, line 11: two nits:
- `executors` => `operations`, these are definitions of individual operations that an executor can run
- I think using a `Map` would be preferable for readability (`if (operations.has(operationName)) { ... }`)

integrations/node/workload-executor.js
```diff
@@ -0,0 +1,100 @@
...
+  find: (collection, args) =>
+    collection.find(args.filter || {}, omit(args, ['filter'])).toArray(),
```
mbroadst, line 13: seems like we would want some more validation here, in case someone fat fingered an operation definition

integrations/node/workload-executor.js
```diff
@@ -0,0 +1,100 @@
...
+
+  updateOne: (collection, args) =>
+    collection.updateOne(args.filter, args.update)
+};
+
+async function main(uri, spec) {
+  const runOperations = makeRunOperations(spec);
+
+  const client = new MongoClient(uri);
+  await client.connect();
```
mbroadst, line 26: just in terms of cleanliness, I think you want to wrap this entire `main` in a try/catch/finally, to ensure that the client is always closed:
```
async function main(..) {
  cons client = new MongoClient(...);

  try {
    await client.connect();
    ... other stuff ...
  } catch (err) { 
    ... handle error ...
  } finally {
    await client.close();
  }

  process.exit(0);
}
```

integrations/node/workload-executor.js
emadum: I removed the `forceServerObjectId` and I'm no longer getting the errors in my evergreen runs; not sure what was going on there, but I'll update if they pop up again.

mbroadst:
@emadum I think this is just waiting on you to resolve my open comments

integrations/node/workload-executor.js
emadum: That timeout actually carried over from the original sample code you sent me, so I wasn't 100% sure what it was for, but it looked like it was there to throttle empty requests. I've renamed it to `NOOP_DELAY`

integrations/node/workload-executor.js
```diff
@@ -5,38 +5,48 @@ const fs = require('fs');
...
-  find: (collection, args) =>
-    collection.find(args.filter || {}, omit(args, ['filter'])).toArray(),
-
-  insertOne: (collection, args) =>
-    collection.insertOne(args.document, { forceServerObjectId: true }),
-
-  updateOne: (collection, args) =>
-    collection.updateOne(args.filter, args.update)
-};
+const operations = new Map();
```
mbroadst, line 13: fwiw you can also do something like:
```
const operations = new Map([
  [
    'find',
    (collection, args) => collection.find(args.filter || {}, omit(args, ['filter'])).toArray()
  ]
]);
```

PR #78, opened 2020-06-24
CSHARP-2515: Test Planned Maintenance Scenarios in Drivers

vincentkam:
First cut at dotnet integration (race condition still extant)
https://spruce.mongodb.com/version/5ef3b7100ae6067429ce7311/tasks
Native workload executor PR: https://github.com/vincentkam/mongo-csharp-driver/pull/119
https://jira.mongodb.org/browse/CSHARP-2515
Edit: As noted in the jira ticket, we will be ignoring the race condition for now

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -181,6 +181,19 @@ axes:
           DRIVER_DIRNAME: "ruby"
           DRIVER_REPOSITORY: "https://github.com/mongodb/mongo-ruby-driver"
           DRIVER_REVISION: "master"
+      - id: dotnet-csharp2515
+        display_name: "dotnet (csharp2515)"
+        variables:
+          DRIVER_DIRNAME: "dotnet"
+          DRIVER_REPOSITORY: "https://github.com/vincentkam/mongo-csharp-driver"
+          DRIVER_REVISION: "csharp2515"
+          ASTROLABE_EXECUTOR_STARTUP_TIME: 40
+          DOTNET_CLI_HOME: "Z:/"
+          TMP: "Z:/"
+          TEMP: "Z:/"
+          NUGET_PACKAGES: "Z:/"
+          NUGET_HTTP_CACHE_PATH: "Z:/"
+          APPDATA: "Z:/"
 
   # The 'platform' axis specifies the evergreen host distro to use.
   # Platforms MUST specify the PYTHON3_BINARY variable as this is required to install astrolabe.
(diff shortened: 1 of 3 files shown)


.evergreen/config.yml
```diff
@@ -181,6 +181,19 @@ axes:
           DRIVER_DIRNAME: "ruby"
           DRIVER_REPOSITORY: "https://github.com/mongodb/mongo-ruby-driver"
           DRIVER_REVISION: "master"
+      - id: dotnet-csharp2515
+        display_name: "dotnet (csharp2515)"
+        variables:
+          DRIVER_DIRNAME: "dotnet"
+          DRIVER_REPOSITORY: "https://github.com/vincentkam/mongo-csharp-driver"
+          DRIVER_REVISION: "csharp2515"
```
vincentkam, line 189: I will change these lines once https://github.com/vincentkam/mongo-csharp-driver/pull/119 is approved and has been merged to master and pushed upstream

.evergreen/config.yml
```diff
@@ -181,6 +181,19 @@ axes:
...
+          ASTROLABE_EXECUTOR_STARTUP_TIME: 40
```
vincentkam, line 190: 40 seconds was needed when the bash script was using `dotnet run` instead of running the published dll/executable. Since we're no longer using `dotnet run` for now, we could perhaps try to find a smaller value for this, but it does no harm to have this value other than slowing the test runs down a bit.

prashantmital: Sounds fine to me to leave it at 40.

integrations/dotnet/workload-executor
```diff
@@ -0,0 +1,63 @@
...
+PLATFORM="" # partial workaround for https://github.com/dotnet/sdk/issues/12159
+export PLATFORM
+
+# Invoke the workload executor as a background process and store its process ID as NATIVE_WORKLOAD_EXECUTOR_PID
+#     We don't need to build or restore because install-driver.sh handles that
+#     We can't use dotnet run --no-build --no-restore due to https://github.com/dotnet/runtime/issues/38274
+#     Addditionally, we must use the published version to bypass a dependency resolution issue with .NET Core 1.1
+if [[ "$FRAMEWORK" == "net452" ]]; then
+    ./mongo-csharp-driver/tests/AstrolabeWorkloadExecutor/bin/Debug/"${FRAMEWORK}"/publish/workload-executor.exe \
+        "${CONNECTION_STRING}" "${WORKLOAD_SPEC}" &
```
prashantmital, line 55: There is no need for running this as background processes so we can remove the trailing `&`.

integrations/dotnet/workload-executor
```diff
@@ -0,0 +1,63 @@
...
+else
+    dotnet mongo-csharp-driver/tests/AstrolabeWorkloadExecutor/bin/Debug/"${FRAMEWORK}"/publish/workload-executor.dll \
```
prashantmital, line 57: Same comment as above about running this as a background process.

integrations/dotnet/workload-executor
```diff
@@ -0,0 +1,63 @@
...
+        "${CONNECTION_STRING}" "${WORKLOAD_SPEC}" &
+fi
+export NATIVE_WORKLOAD_EXECUTOR_PID=$!
+
+# Wait for a state change in $NATIVE_WORKLOAD_EXECUTOR_PID (without this the foreground process would return)
+wait $NATIVE_WORKLOAD_EXECUTOR_PID
```
prashantmital, line 63: We don't need this wait once the executor is run as a foreground process.

integrations/dotnet/workload-executor
```diff
@@ -0,0 +1,63 @@
...
+else
+  RESULTS_DIR="$(pwd)"
+fi
+
+# Set a trap for the INT
+# - wait for a state change in child process bearing ID $NATIVE_WORKLOAD_EXECUTOR_PID
+# - exit with the same status/code as the exit code of the call to wait (which returns the exit code of $NATIVE_WORKLOAD_EXECUTOR_PID)
+# Note that calls to wait return immediately if the child has already changed state
+
+trap 'echo You have activated my trap card; wait $NATIVE_WORKLOAD_EXECUTOR_PID; exit $?' INT
```
prashantmital, line 41: The trap can be removed since the executor will now be run in the foreground. See comments below.

prashantmital:
Superseded by #86

PR #81, opened 2020-07-08
Fix node integration

emadum:
### Description

- Fixes a bug with duplicate IDs on insert by passing a copy of operation arguments instead of a reference
- Remove node_carbon runtime (no longer supported in driver version `4.0`/`master`)
- Print errors to stdout for easier future debugging

[Patch results](https://spruce.mongodb.com/version/5f10bd6da4cf472226e658b9/tasks)

### DIFF
--- a/integrations/node/install-driver.sh
+++ b/integrations/node/install-driver.sh
@@ -47,3 +47,7 @@ npm install yargs lodash.omit
 cd ${PROJECT_DIRECTORY}
 npm install
 
+# compile typescript if necessary
+if [ $DRIVER_REVISION == "master" ]; then
+    npm run build:ts
+fi
--- a/integrations/node/install-driver.sh
+++ b/integrations/node/install-driver.sh
@@ -46,8 +46,3 @@ npm install yargs lodash.omit
 # install node dependencies
 cd ${PROJECT_DIRECTORY}
 npm install
-
-# compile typescript if necessary
-if [ $DRIVER_REVISION == "master" ]; then
-    npm run build:ts
-fi
--- a/integrations/node/workload-executor
+++ b/integrations/node/workload-executor
@@ -9,11 +9,6 @@ PATH=$(echo "$PATH" | sed -e 's/:\/opt\/node\/bin$//')
 PROJECT_DIRECTORY="$(pwd)/node-mongodb-native"
 NODE_ARTIFACTS_PATH="${PROJECT_DIRECTORY}/node-artifacts"
 
-# master now uses typescript, point to compiled js in lib/
-if [ $DRIVER_REVISION == "master" ]; then
-    PROJECT_DIRECTORY="${PROJECT_DIRECTORY}/lib"
-fi
-
 export PROJECT_DIRECTORY
 
 export NVM_DIR="${NODE_ARTIFACTS_PATH}/nvm"
--- a/integrations/node/workload-executor
+++ b/integrations/node/workload-executor
@@ -6,9 +6,15 @@ set -o errexit  # Exit the script with error if any of the commands fail
 # make sure there are no node/npm in path
 PATH=$(echo "$PATH" | sed -e 's/:\/opt\/node\/bin$//')
 
-PWD=$(pwd)
-PROJECT_DIRECTORY="$PWD/node-mongodb-native"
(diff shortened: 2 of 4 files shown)


emadum:
@mbroadst Do you think it'd make sense to add the `3.5` and `3.6` branches into this?

emadum:
With `3.5` and `3.6`: [Patch results](https://spruce.mongodb.com/version/5f07178432f4176793384a46/tasks)

integrations/node/install-driver.sh
```diff
@@ -47,3 +47,7 @@ npm install yargs lodash.omit
 cd ${PROJECT_DIRECTORY}
 npm install
 
+# compile typescript if necessary
```
mbroadst, line 50: this is the change to the driver we discussed yesterday that I think would obviate this step:
```
diff --git a/package.json b/package.json
index 32df1e850..a5896ab70 100644
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "check:ts": "tsc --noEmit",
     "check:atlas": "mocha --config \"test/manual/mocharc.json\" test/manual/atlas_connectivity.test.js",
     "check:ocsp": "mocha --config \"test/manual/mocharc.json\" test/manual/ocsp_support.test.js",
-    "prepublishOnly": "npm run build:ts",
+    "prepare": "npm run build:ts",
     "release": "standard-version -i HISTORY.md",
     "test": "npm run check:lint && npm run check:test"
   },
```

integrations/node/workload-executor
```diff
@@ -6,10 +6,16 @@ set -o errexit  # Exit the script with error if any of the commands fail
 # make sure there are no node/npm in path
 PATH=$(echo "$PATH" | sed -e 's/:\/opt\/node\/bin$//')
 
-PWD=$(pwd)
-PROJECT_DIRECTORY="$PWD/node-mongodb-native"
+PROJECT_DIRECTORY="$(pwd)/node-mongodb-native"
 NODE_ARTIFACTS_PATH="${PROJECT_DIRECTORY}/node-artifacts"
 
+# master now uses typescript, point to compiled js in lib/
+if [ $DRIVER_REVISION == "master" ]; then
```
mbroadst, line 13: I think you could also remove this special casing with the change above

integrations/node/workload-executor.js
```diff
@@ -37,6 +37,7 @@ async function main(uri, spec) {
       try {
         await runOperations(client, results);
       } catch (err) {
+        console.log({ err });
```
mbroadst, line 40: I think we actually want `console.dir(err)` here instead, unless you think seeing the wrapping `{ err: ... }` is useful for debugging?

integrations/node/workload-executor.js
```diff
@@ -72,7 +73,10 @@ function makeRunOperations(spec) {
             ++results.numFailures;
           }
         })
-        .catch(() => (++results.numErrors)),
+        .catch(err => {
+          console.log({ err });
```
mbroadst, line 77: same comment as above

integrations/node/workload-executor.js
```diff
@@ -87,7 +91,7 @@ function runOperation(op) {
     throw new Error(`Unsupported operation: ${op.name}`);
   }
   const operation = operations.get(op.name);
-  return operation(object, op.arguments);
+  return operation(object, JSON.parse(JSON.stringify(op.arguments)));
```
mbroadst, line 94: Let's wrap this in a `try/catch` to make it very obvious to future maintainers why this fails if it does

integrations/node/install-driver.sh
nbbeeken: I think the prepare script will have us covered, the npm install above will run the prepare script so the driver will be compiled if it's from master.

mbroadst:
@emadum what's going on with this PR?

emadum:
> @emadum what's going on with this PR?

Just need to address your feedback regarding the `prepare` npm script - will do so today.

integrations/node/install-driver.sh
emadum: Updated, PTAL

emadum:
[EVG Patch](https://spruce.mongodb.com/version/5f10bd6da4cf472226e658b9/tasks)

Issue #82, opened 2020-08-06
Make validation a bit more forgiving


jyemin:
The `validate-workload-executor` checks the count of ops against the expected count.  It should allow one more operation than expected, so that workload executors can write the results file in the signal handler without waiting for the main loop to actually exit.

prashantmital:
I think allowing +-1 margin of error here is perfectly reasonable and probably required due to the limitations on signal handling in various languages.

PR #85, opened 2020-08-06
Add Java driver workload executor

jyemin:
* Add install script
* Add workload executor script
* Add Evergreen configuration

See https://jira.mongodb.org/browse/JAVA-3201 for a link to the mongo-java-driver repository commit on which this depends

Patch build: https://spruce.mongodb.com/version/5f2c5b8e32f41766c440b72c/tasks

Note that the validate-workload-executor task fails due to https://github.com/mongodb-labs/drivers-atlas-testing/issues/82

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@
 astrolabe.egg-info/
 results.json
 xunit-output/
+*~
+*.iml
--- a/integrations/java/install-driver.sh
+++ b/integrations/java/install-driver.sh
@@ -2,8 +2,6 @@
 
 set -o errexit
 
-export JAVA_HOME="/opt/java/jdk11"
-
 cd mongo-java-driver || exit
 
 ./gradlew --info driver-workload-executor:compileJava
--- a/integrations/java/install-driver.sh
+++ b/integrations/java/install-driver.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+set -o errexit
+
+export JAVA_HOME="/opt/java/jdk11"
+
+cd mongo-java-driver || exit
+
+./gradlew --info driver-workload-executor:compileJava
--- a/integrations/java/workload-executor
+++ b/integrations/java/workload-executor
@@ -2,8 +2,6 @@
 
 set -o errexit
 
-export JAVA_HOME="/opt/java/jdk11"
-
 OUTPUT_DIRECTORY=$PWD
 
 cd mongo-java-driver || exit
--- a/integrations/java/workload-executor
+++ b/integrations/java/workload-executor
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+set -o errexit
+
+export JAVA_HOME="/opt/java/jdk11"
+
+OUTPUT_DIRECTORY=$PWD
+
+cd mongo-java-driver || exit
+
+./gradlew --info  -Dorg.mongodb.workload.output.directory="$OUTPUT_DIRECTORY" -Dorg.mongodb.test.uri="$1" -Dorg.mongodb.workload.spec="$2" driver-workload-executor:run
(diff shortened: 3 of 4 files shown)


.gitignore
```diff
@@ -3,3 +3,5 @@
 astrolabe.egg-info/
 results.json
 xunit-output/
+*~
```
jyemin, line 6: Drive by fix in separate commit since I almost committed files with these extensions by accident. .iml is a Jet Brains thing FYI.

integrations/java/install-driver.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -o errexit
+
+export JAVA_HOME="/opt/java/jdk11"
+
+cd mongo-java-driver || exit
+
+./gradlew --info driver-workload-executor:compileJava
```
jyemin, line 9: This will end up downloading a lot of stuff from Maven Central as well as compiling the whole driver.  Typically takes about 2 minutes, so better to do it here than in the executor.

integrations/java/workload-executor
```diff
@@ -0,0 +1,11 @@
...
+
+set -o errexit
+
+export JAVA_HOME="/opt/java/jdk11"
+
+OUTPUT_DIRECTORY=$PWD
+
+cd mongo-java-driver || exit
+
+./gradlew --info  -Dorg.mongodb.workload.output.directory="$OUTPUT_DIRECTORY" -Dorg.mongodb.test.uri="$1" -Dorg.mongodb.workload.spec="$2" driver-workload-executor:run
```
jyemin, line 11: As discussed: running it in the foreground works just fine

integrations/java/workload-executor
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -o errexit
+
+export JAVA_HOME="/opt/java/jdk11"
+
+OUTPUT_DIRECTORY=$PWD
+
+cd mongo-java-driver || exit
```
prashantmital, line 9: Can you explain why this is needed?

prashantmital: Can this be changed as follows:

```
cd mongo-java-driver && ./gradlew --info  -Dorg.mongodb.workload.output.directory="$OUTPUT_DIRECTORY" -Dorg.mongodb.test.uri="$1" -Dorg.mongodb.workload.spec="$2" driver-workload-executor:run
```

integrations/java/workload-executor
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -o errexit
+
+export JAVA_HOME="/opt/java/jdk11"
```
prashantmital, line 5: I think you should define this in the `runtime` axis to make it easy to test using other JDK versions going-forward. Both node and Python already do this.

integrations/java/install-driver.sh
prashantmital: Ack.

.gitignore
prashantmital: Ack.

integrations/java/workload-executor
jyemin: Why is the `exit` needed?  It's just a sanity check in case the directory isn't there.

integrations/java/workload-executor
jyemin: I don't see a need to test any other JDKs.  Can we leave it here until we have the need?

jyemin: If I did do it that way, would the variable I define there be accessible in this script?  e.g. 

    export JAVA_HOME="/opt/java/$JDK. 

where the runtime access had

   JDK: "jdk11"

?

prashantmital: Yes. You could actually just define `JAVA_HOME` in its entirety in the `runtime` axis so this `export` statement wouldn't be needed at all.

jyemin: Ok, you convinced me.  Done.

PR #87, opened 2020-08-19
Add go driver integration

iwysiu:


integrations/go/workload-executor.go
```diff
@@ -0,0 +1,315 @@
...
+	}
+	return false, errors.New("unrecognized object: " + op.Name)
+}
+
+func main() {
+	connstring := os.Args[1]
+	workloadSpec := os.Args[2]
+
+	var workload driverWorkload
+
```
divjotarora, line 234: nit: remove this blank line

integrations/go/workload-executor.go
```diff
@@ -0,0 +1,315 @@
...
+	results := struct {
+		NumErrors    int `json:"numErrors"`
+		NumFailures  int `json:"numFailures"`
+		NumSuccesses int `json:"numSuccesses"`
+	}{}
+
+	done := make(chan struct{})
+
+	go func() {
```
divjotarora, line 259: Can you add a comment explaining what this goroutine is doing? Also, are we always guaranteed to get a SIGTERM?

integrations/go/workload-executor.go
```diff
@@ -0,0 +1,315 @@
...
+
+func main() {
+	connstring := os.Args[1]
+	workloadSpec := os.Args[2]
+
+	var workload driverWorkload
+
+	err := bson.UnmarshalExtJSONWithRegistry(specTestRegistry,[]byte(workloadSpec), false, &workload)
+	if err != nil {
+		str := "failed to unmarshal data: " + err.Error()
```
divjotarora, line 237: nit: You usually don't need to add contexts to panics because the panic output will have a stack trace with the line number. If you feel that these are helpful, I'm fine leaving them in, but would prefer using `fmt.Sprintf` here rather than string concatenation to avoid any weird spacing issues if we make these messages more complex in the future.

integrations/go/workload-executor.go
```diff
@@ -0,0 +1,315 @@
...
+		}
+		path, _ := os.Getwd()
+		err = ioutil.WriteFile(path+"/results.json", data, 0644)
+		if err != nil {
+			str := "write to file failed: " + err.Error()
+			panic(str)
+		}
+	}()
+
+	for {
```
divjotarora, line 292: Why do we need a nested loop here? Is this equivalent to:

```
// do opportunistic check for done
select {
     case <-done:
         return
     default:
}

// Run ops, checking for done before each op
for _, operation : = range operations {
    // same select as above with done/default cases
    // runOp
}

// Wait for done before returning
// select with single done case, no default

File: integrations/go/workload-executor.go (hunk 1 above)
Comment (iwysiu) [reply integrations/go/workload-executor.go]: done

File: integrations/go/workload-executor.go (hunk 3 above)
Comment (iwysiu) [reply integrations/go/workload-executor.go]: done

File: integrations/go/workload-executor.go (hunk 2 above)
Comment (iwysiu) [reply integrations/go/workload-executor.go:258]: done. and yup, we repeat the operations until astrolabe sends a termination signal

File: integrations/go/workload-executor.go (hunk 4 above)
Comment (iwysiu) [reply integrations/go/workload-executor.go:291]: i cleaned it up, but we do want it to repeat the operations from the top until astrolabe ends it

File: integrations/go/workload-executor.go [hunk 5]
```diff
@@ -0,0 +1,314 @@
...
+	elems, _ := args.Elements()
+	for _, elem := range elems {
+		key := elem.Key()
+		val := elem.Value()
+
+		switch key {
+		case "document":
+			doc = val.Document()
+		default:
+			return nil, errors.New("unrecognized insertOne option: " + key)
```
divjotarora, line 53: My understanding is that we'd end up recording this as an error so the `numErrors` field would be incremented and the test would fail. If this is the case, though, we lose the information about why it errored and we also risk having a false positive if the operation was actually supposed to error for some other reason. All of the `unrecognized...` default cases block us from even running the operation, so can we just change them to panic instead?

integrations/go/workload-executor.go
```diff
@@ -0,0 +1,314 @@
...
+}
+
+func main() {
+	connstring := os.Args[1]
+	workloadSpec := os.Args[2]
+
+	var workload driverWorkload
+	err := bson.UnmarshalExtJSONWithRegistry(specTestRegistry, []byte(workloadSpec), false, &workload)
+	if err != nil {
+		panic(err.Error())
```
divjotarora, line 237: nit: change to `panic(err)`

integrations/go/workload-executor.go
```diff
@@ -0,0 +1,314 @@
...
+		close(done)
+	}()
+
+	_ = coll.Drop(context.Background())
+
+	// insert testdata
+	if len(workload.TestData) > 0 {
+		_, err := coll.InsertMany(context.Background(), workload.TestData)
+		if err != nil {
+			str := fmt.Sprintf("inserting testData failed: %s", err.Error())
```
divjotarora, line 272: nit: Use the `%v` directive: `fmt.Sprintf("inserting testData failed: %v", err)`.

integrations/go/workload-executor.go
```diff
@@ -0,0 +1,306 @@
...
+func verifyCursorResult(cur *mongo.Cursor, result interface{}) bool {
+	if result == nil {
+		return true
+	}
+
+	if cur == nil {
+		return false
+	}
+
+	defer func() {
```
divjotarora, line 162: We should be checking the error from `cur.Close` in the defer. We can probably just panic if it errors.

integrations/go/workload-executor.go
```diff
@@ -0,0 +1,306 @@
...
+		for _, operation := range workload.Operations {
+			select {
+			case <-done:
+				return
+			default:
+				pass, err := runOperation(coll, operation)
+				switch {
+				case err != nil:
+					results.NumErrors++
+				case !pass:
```
divjotarora, line 298: nit: super minor, but we usually prefer checking positives instead of negatives. Can we change this to `case pass:` and have it increment `NumSuccesses` instead?

iwysiu: done

integrations/go/workload-executor.go
iwysiu: done

prashantmital (approved):
LGTM

integrations/go/workload-executor.go
```diff
@@ -0,0 +1,309 @@
...
+		NumFailures  int `json:"numFailures"`
+		NumSuccesses int `json:"numSuccesses"`
+	}{}
+
+	done := make(chan struct{})
+
+	// Waits for the termination signal from astrolabe and terminates the operation loop
+	go func() {
+		c := make(chan os.Signal, 1)
+		signal.Notify(c, os.Interrupt, syscall.SIGTERM)
```
kevinAlbs, line 266: To check, this does not run on Windows, right? Otherwise, this may need to account for `CTRL_CLOSE_EVENT` as described in the Astrolabe docs https://mongodb-labs.github.io/drivers-atlas-testing/spec-workload-executor.html#behavioral-description

> The choice of termination signal used by astrolabe varies by platform. SIGINT 1 is used as the termination signal on Linux and OSX, while CTRL_BREAK_EVENT 2 is used on Windows.

integrations/go/workload-executor.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+	return false, errors.New("unrecognized collection operation: " + op.Name)
+}
+
+func runOperation(coll *mongo.Collection, op *operation) (bool, error) {
+	// execute the command on the given object
+	if op.Object == "collection" {
+		return executeCollectionOperation(coll, op)
+	}
+	return false, errors.New("unrecognized object: " + op.Name)
```
kevinAlbs, line 233: According to https://mongodb-labs.github.io/drivers-atlas-testing/spec-test-format.html#test-scenario-format-specification object could also be "database". Though I assume if this is passing, perhaps that is not the case.

I don't see any mention of "database" in the operations of https://github.com/mongodb-labs/drivers-atlas-testing/tree/master/tests. But perhaps this should panic instead of returning an error? A hard crash may be easier to interpret if "database" was added to the format later. Currently I think this error message won't be printed anywhere.

integrations/go/workload-executor.go
```diff
@@ -0,0 +1,309 @@
...
+		res, err := executeInsertOne(coll, op.Arguments)
+		return verifyInsertOneResult(res, op.Result), err
+	case "find":
+		cursor, err := executeFind(coll, op.Arguments)
+		return verifyCursorResult(cursor, op.Result), err
+	case "updateOne":
+		res, err := executeUpdateOne(coll, op.Arguments)
+		return verifyUpdateResult(res, op.Result), err
+	}
+	return false, errors.New("unrecognized collection operation: " + op.Name)
```
kevinAlbs, line 225: Minor, but I'd argue for changing this to a "panic". I think the error message is not printed currently. If the test format changes, it'd be a clearer error.

kevinAlbs (commented):
Looks great! Just have a couple questions.

integrations/go/workload-executor.go
prashantmital: The idea behind allowing object to be "database" was that we _could_ write tests that ran operations on Database objects but none of the existing planned maintenance tests do this. Eventually, much of the format used for these tests will match what comes out of the Unified Test Format project so I don't see any reason to add support for database at this point in time.

integrations/go/workload-executor.go
divjotarora: Good catch. I left a similar comment earlier about functions like `executeInsertOne` (https://github.com/mongodb-labs/drivers-atlas-testing/pull/87#discussion_r473237767). To generalize, we should only return errors once we can actually call driver functions. If there are any errors during setup for an operation, those should panic because they signal a deterministic flaw in the test runner, not a non-deterministic runtime error.

prashantmital: This is an interesting discussion. Having the executor panic seems like the right thing to do here. The validator example is a bit contrived because like @iwysiu said, it uses a non-existent operation to check that the executor increments the error counter correctly. It would be much more realistic to use the executor to run an operation that is guaranteed to fail (most likely using a failPoint) to verify this behavior. I can make a quick PR to make this change if you'd rather panic here than return an error.

iwysiu: @prashantmital I think that would make a lot of sense. Thanks!

kevinAlbs (approved):
LGTM!

PR #88, opened 2020-08-26
Make workload executor validation more permissive 

### LINKED ISSUE #82: Make validation a bit more forgiving

prashantmital:
Closes #82


prashantmital:
Validation passes on all drivers with the more permissive assertion logic.

PR #89, opened 2020-08-26
Bash scripts wrapping native workload executors need not run them as background processes

### LINKED ISSUE #83: Bash scripts wrapping native workload executors need not run them as background processes
Originally, the bash wrapper for native workload executors ran the native script/executable as a background process so that the script's exit code upon termination by astrolabe could be set to be the same as the exit code of the native workload executor. This was needed because:

* bash scripts that wrapped a native executor always gave a non zero code if they handled a SIGINT without setting a trap
* astrolabe relied on the exit code to ascertain test-run success/failure

Astrolabe now uses execution statistics to determine test success/failure so we no longer need to do this. This significantly simplifies the workload executor wrapper script.

As part of this ticket, we should:

- [ ] Update the documentation/specification
- [ ]  Update the Python workload executor wrapper since this is referenced in the documentation as an example of how to wrap a native executor in a bash script

prashantmital:
Closes #83

### DIFF
--- a/integrations/python/pymongo/workload-executor
+++ b/integrations/python/pymongo/workload-executor
@@ -1,14 +1,5 @@
 #!/bin/bash
 
-# Set a trap for the INT signal:
-# - wait for a state change in child process bearing ID $PID
-# - exit with the same status/code as the exit code of the call to wait (which returns the exit code of $PID)
-# Note that calls to wait return immediately if the child has already changed state
-trap 'wait $PID; exit $?' INT
+set -o errexit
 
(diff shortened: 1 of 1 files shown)


prashantmital:
Seems like we don't need to do anything here other than modifying the Python workload executor wrapper script itself. The documentation/spec just refers to this spec and makes no mention of running the workload executors as background processes and existing integrations already adhere to this pattern.

prashantmital:
Java validation failures will be resolved by #88

PR #90, opened 2020-08-27
Document clearly that workload executors MUST ignore the testData key in the workload specification

### LINKED ISSUE #84: Document that workload executors MUST ignore the testData key in the workload executor spec
This is currently only documented in the YAML test format specification. We should document it in the workload execution spec (and add a comment about it to the workload executor pseudocode) because users are likelier to be reading that document when implementing their executors.

prashantmital:
Closes #84


prashantmital:
This is something that came up during the Java driver's implementation. Ideally, this key would not be part of the `driverWorkload` section of the test spec, but we are stuck with that for now. 

In addition to the documentation changes, I can remove this key from the workload spec that is passed by `astrolabe` to the workload executors to ensure that none of them are inadvertently doing this.

PR #91, opened 2020-09-11
Change schema to support multiple operations


p-mongo:


mbroadst:
@p-mongo can you please provide a description for this change? There isn't a ticket associated either, so I'm not sure how we're supposed to review this.

astrolabe/runner.py
```diff
@@ -161,34 +157,35 @@ def run(self, persist_cluster=False, startup_time=1):
             driver_workload=self.spec.driverWorkload,
             startup_time=startup_time)
 
-        # Step-3: begin maintenance routine.
-        final_config = self.spec.maintenancePlan.final
-        cluster_config = final_config.clusterConfiguration
-        process_args = final_config.processArgs
+        for operation in self.spec.operations:
+            # Step-3: begin maintenance routine.
+            final_config = operation.setClusterConfiguration
```
prashantmital, line 162: We should swap out everything inside this `for ...` with a single call to a utility function like `run_maintenance_operation`. That way there is an easy to find place to add maintenance operations.

prashantmital (commented):
Shouldn't we wait until the move to the operations-based maintenance plan is approved in the design phase?

p-mongo:
I will open a new PR when the design is approved.

PR #94, opened 2020-09-28
Use cloud-dev to run Atlas Planned Maintenance tests

prashantmital:


.evergreen/config.yml
```diff
@@ -78,6 +78,8 @@ functions:
           EVERGREEN_BUILD_ID: ${build_id}
           ATLAS_API_USERNAME: ${atlas_key}
           ATLAS_API_PASSWORD: ${atlas_secret}
+          ATLAS_API_BASE_URL: ${atlas_url}
+          ATLAS_ORGANIZATION_NAME: ${atlas_organization}
```
p-mongo, line 82: Why is the org name needed?

prashantmital: Because the default value in `astrolabe` no longer matches the name of the org to which the cloud-dev API key belongs. Instead of changing the default (set [here](https://github.com/mongodb-labs/drivers-atlas-testing/blob/96ceb70cb9827479fd35c0199cc64d29fed574a3/astrolabe/configuration.py#L20)) I instead started using the envvar to configure this value in case it changes again in the future. The envvar to use can be ascertained/modified using [this code](https://github.com/mongodb-labs/drivers-atlas-testing/blob/96ceb70cb9827479fd35c0199cc64d29fed574a3/astrolabe/configuration.py#L30).

prashantmital: Note that you can also get this information by running `astrolabe info default-values` (displays config opt name and corresponding default value) and `astrolabe info environment-variables` (displays envvars that can set to configure various opts).

PR #98, opened 2021-01-05
DRIVERS-828 comprehensive Atlas testing

p-mongo:
https://jira.mongodb.com/browse/DRIVERS-828


astrolabe/runner.py
```diff
@@ -214,6 +256,11 @@ def run(self, persist_cluster=False, startup_time=1):
             # is only visible for failed tests.
 
         LOGGER.info("Workload Statistics: {}".format(stats))
+        
+        #LOGGER.info("Waiting 5 minutes for Atlas logs to become available")
+        #sleep(5*60)
+
```
alcaeus, line 262: This looks like debug code that should be removed.
```suggestion
```

alcaeus (commented):
Nit about debug code being left in. I'll leave the Python code review to someone that knows Python, but spec changes LGTM.

astrolabe/runner.py
p-mongo: Actually this was needed prior to the current method of log retrieval, but currently this code has no purpose.

astrolabe/cli.py
```diff
@@ -506,5 +537,39 @@ def validate_workload_executor(workload_executor, startup_time,
...
+@click.pass_context
+def stats(ctx):
+    with open('results.json', 'r') as fp:
+        stats = json.load(fp)
+    with open('events.json', 'r') as fp:
+        events = json.load(fp)
+    
+    import numpy
+    
+    command_events = events['commands']
```
prashantmital, line 550: Can you refactor this into a function in the `astrolabe.commands` submodule and call it here instead? The `astrolable.cli` submodule should just have the code that parses the command-line input and dispatches the work.

astrolabe/runner.py
```diff
@@ -91,29 +93,31 @@ def is_cluster_state(self, goal_state):
...
+            process_args, expected_configuration.processArgs)
 
-    def initialize(self):
+    def initialize(self, no_create=False):
         """
         Initialize a cluster with the configuration required by the test
         specification.
         """
+        
+        self.start_time = _time.time()
```
prashantmital, line 113: Where is this being used? Also, you should use `astrolabe.utils.Timer` for timing operations - it exposes a simpler interface and uses a monotonic timer to nullify jumps in wall-clock time (which is what is returned by `_time.time`)

astrolabe/runner.py
```diff
@@ -91,29 +93,31 @@ def is_cluster_state(self, goal_state):
...
+        
+        if no_create:
+            return
```
prashantmital, line 116: Shouldn't you at least check that a cluster of the expected name exists before returning? Also, I think you should emit a `LOGGER.info` event before returning.

astrolabe/runner.py
```diff
@@ -163,34 +167,72 @@ def run(self, persist_cluster=False, startup_time=1):
...
-        selector = BooleanCallablePoller(
-            frequency=self.config.polling_frequency,
-            timeout=self.config.polling_timeout)
-        LOGGER.info("Waiting for cluster maintenance to complete")
-        selector.poll([self], attribute="is_cluster_state", args=("IDLE",),
-                      kwargs={})
-        self.verify_cluster_configuration_matches("final")
-        LOGGER.info("Cluster maintenance complete")
+        for operation in self.spec.operations:
+            if hasattr(operation, 'setClusterConfiguration'):
```
prashantmital, line 171: Since a single entry in `spec.operations` can only have one operation, I suggest the following refactor to make this all a little easier to read:

```
op_name, op_spec = next(iteritems(operation))

if op_name == 'setClusterConfiguration':
    cluster_config = op_spec.clusterConfiguration
    process_args = op_spec.processArgs
    ...
elif op_name == 'testFailover':
    ...
```

astrolabe/runner.py
```diff
@@ -163,34 +167,72 @@ def run(self, persist_cluster=False, startup_time=1):
...
+                if cluster_config:
+                    LOGGER.info("Pushing cluster configuration update")
+                    self.cluster_url.patch(**cluster_config)
+
+                if process_args:
+                    LOGGER.info("Pushing process arguments update")
+                    self.cluster_url.processArgs.patch(**process_args)
+
+                # Sleep before polling to give Atlas time to update cluster.stateName.
+                sleep(3)
```
prashantmital, line 189: I think instead of waiting here, we should introduce a wait at the start of `wait_for_idle()`. That will provide a better abstraction and will prevent these somewhat random sleep times from being scattered all over the codebase.

astrolabe/runner.py
```diff
@@ -163,34 +167,72 @@ def run(self, persist_cluster=False, startup_time=1):
...
+                _time.sleep(operation['sleep'])
+                
+            if hasattr(operation, 'waitForIdle'):
+                self.wait_for_idle()
+                
+            if hasattr(operation, 'restartVms'):
+                url = "/api/private/nds/groups/%s/clusters/%s/reboot" % (self.project.id, self.cluster_name)
+                self.admin_client.request('POST', url)
+                
+                sleep(5)
```
prashantmital, line 211: Same comment as above about sleeping in `wait_for_idle` instead of before calling it. If you'd like, it can be a kwarg passed to `wait_for_idle`, e.g. `wait_for_idle(sleep=5)` will sleep 5 seconds before polling server state.

astrolabe/runner.py
```diff
@@ -163,34 +167,72 @@ def run(self, persist_cluster=False, startup_time=1):
...
+                self.wait_for_idle()
+                
+            if hasattr(operation, 'assertPrimaryRegion'):
+                region = operation['assertPrimaryRegion']
+                
+                cluster_config = self.cluster_url.get().data
+                deadline = monotonic() + 90
```
prashantmital, line 218: Does 90 come from the spec? What is the basis for this choice? I think we need to add a comment here describing where this number came from.

Also, we should once again use `astrolabe.utils.Timer` here as follows:

```
from astrolabe.utils import Timer

timer = Timer()
timer.start()

while True:
    # Do stuff
    ...
    if timer.elapsed > 90:
        raise ...
```

astrolabe/runner.py
```diff
@@ -163,34 +167,72 @@ def run(self, persist_cluster=False, startup_time=1):
...
+                
+                while True:
+                    mc = MongoClient(cluster_config['connectionStrings']['standard'], username='atlasuser', password='mypassword123')
```
prashantmital, line 221: Why does this need a new MongoClient? Why can we not use one of the clients attached to the context for this operation?

atlasclient/client.py
```diff
@@ -218,7 +220,10 @@ def request(self, method, path, **params):
         return self.handle_response(method, response)
 
     def construct_resource_url(self, path):
-        url_template = "{base_url}/v{version}/{resource_path}"
+        if path and path[0] == '/':
+            url_template = 'https://cloud-dev.mongodb.com{resource_path}'
```
prashantmital, line 224: We should use the `base_url` config var to set this instead of hardcoding it... Does the cloud-dev API not use the same convention as the public API?

prashantmital (changes requested):
I haven't had a chance to go through all of the changes yet. Publishing my initial comments so we can start iterating.

.evergreen/config.yml
```diff
@@ -98,7 +100,7 @@ functions:
         working_dir: astrolabe-src
         add_expansions_to_env: true
         command: |
-          astrolabevenv/${PYTHON_BIN_DIR}/astrolabe spec-tests validate-workload-executor -e integrations/${DRIVER_DIRNAME}/workload-executor --connection-string "mongodb://localhost:27017/"
+          astrolabevenv/${PYTHON_BIN_DIR}/astrolabe spec-tests validate-workload-executor -e integrations/${DRIVER_DIRNAME}/workload-executor --connection-string "mongodb://localhost:27017/?serverselectiontimeoutms=10000"
```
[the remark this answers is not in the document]
p-mongo: I intended this to make the tests fail faster when the deployment is not available for whatever reason (most likely user error), though I can't remember what the exact issue was at this point.

astrolabe/runner.py
```diff
@@ -223,17 +267,63 @@ def run(self, persist_cluster=False, startup_time=1):
...
+            rname = data['deploymentItemName']
+            
+        params = dict(
+            resourceName=rname,
+            resourceType=rtype,
+            redacted=True,
+            logTypes=['FTDC','MONGODB'],#,'AUTOMATION_AGENT','MONITORING_AGENT','BACKUP_AGENT'],
+            sizeRequestedPerFileBytes=100000000,
+        )
+        data = self.admin_client.request('POST', 'groups/%s/logCollectionJobs' % (self.project.id,), **params).data
```
prashantmital, line 296: The `AtlasClient` object exposes a fluent API that you should use instead of issuing a raw request here. In this instance, this would look like:

```
data = self.admin_client.groups[self.project.id].logCollectionJobs.post(**params).data
```

atlasclient/client.py
prashantmital: I have opened #99 that makes this change unnecessary.

astrolabe/runner.py
```diff
@@ -223,17 +267,63 @@ def run(self, persist_cluster=False, startup_time=1):
...
+    def wait_for_idle(self):
+        selector = BooleanCallablePoller(
+            frequency=self.config.polling_frequency,
+            timeout=self.config.polling_timeout)
+        LOGGER.info("Waiting for cluster maintenance to complete")
+        selector.poll([self], attribute="is_cluster_state", args=("IDLE",),
+                      kwargs={})
+                      
+    def retrieve_logs(self):
+        data = self.admin_client.request('GET', '/api/private/nds/groups/%s/clusters/%s' % (self.project.id, self.cluster_name)).data
```
prashantmital, line 280: After rebasing on #99, and reverting the changes made in this PR to `AtlasClient.construct_resource_url`, this should simply be:

```
data = self.admin_client.nds.groups[self.project.id].clusters[self.cluster_name].get(api_version='private').data
```

prashantmital (changes requested):

.evergreen/config.yml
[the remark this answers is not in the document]
prashantmital: Ok. We can leave it as-is since this is only for validating the executors themselves and does not impact testing.

astrolabe/cli.py
p-mongo: Moved.

astrolabe/runner.py
p-mongo: This was used for the delay to retrieve Atlas logs using the public API but is not currently used for anything, I'll remove.

astrolabe/runner.py
p-mongo: You are right that the diagnostics would be better if the name was verified. Thinking about this some more, I'll make --no-create create the cluster if it doesn't exist which should make this option more usable without imposing additional runtime cost.

astrolabe/runner.py
p-mongo: Done, thanks for the suggestion.

astrolabe/runner.py
p-mongo: Actually this sleep isn't necessary anymore because the wait is now part of the scenarios since 3 seconds was not always sufficient. I'll remove it.

astrolabe/runner.py
p-mongo: I will move this also to scenario specifications.

astrolabe/runner.py
p-mongo: I think the value was arbitrary but I'll make it overridable in the scenario.

astrolabe/runner.py
p-mongo: I see two other clients in the codebase but neither of them appears to be attached to any contexts, which client did you have in mind exactly?

tests/retryReads-move-sharded.yaml
```diff
@@ -0,0 +1,62 @@
+initialConfiguration:
```
jmikola, line 1: Is there a particular reason some files in this PR use `.yml` but others use `.yaml`? AFAIK, mongodb/specifications uses `.yml` exclusively so perhaps we can stick to that for consistency.

astrolabe/cli.py
```diff
@@ -14,8 +14,9 @@
 
 import logging
 from pprint import pprint
-import unittest
+import unittest, os
```
prashantmital, line 17: Import `os` is unused.

astrolabe/cli.py
```diff
@@ -506,5 +536,11 @@ def validate_workload_executor(workload_executor, startup_time,
         exit(1)
 
 
+@spec_tests.command('stats')
```
prashantmital, line 539: This can simply be `@spec_tests.command()` . You only need to include the name of the command in the call to `command()` if it differs from the name of the method being decorated.

astrolabe/cli.py
```diff
@@ -506,5 +536,11 @@ def validate_workload_executor(workload_executor, startup_time,
...
+@click.pass_context
+def stats(ctx):
```
prashantmital, line 541: Do we call this anywhere or is it a utility for developing/debugging? If it is only a utility, shouldn't we also print out these stats instead of just writing them to a file? (assuming that developing/debugging workflows would usually involve interacting with Astrolabe in a shell)

astrolabe/runner.py
```diff
@@ -163,34 +173,73 @@ def run(self, persist_cluster=False, startup_time=1):
...
+            if op_name == 'assertPrimaryRegion':
+                region = op_spec['region']
+                
+                cluster_config = self.cluster_url.get().data
+                timer = Timer()
+                timer.start()
+                timeout = op_spec.get('timeout', 90)
+                
+                while True:
+                    with mongo_client(self.get_connection_string()) as mc:
```
prashantmital, line 230: Why are we not simply using `self.client` here? We don't seem to modify any of the URI options and directly use the connection string so creating a new client everytime to run this operation appears to be pointless. Can you add the rationale for using a new client as a comment?

astrolabe/validator.py
```diff
@@ -18,33 +18,27 @@
...
     def setUp(self):
         self.client = MongoClient(self.CONNECTION_STRING, w='majority')
-        self.coll = self.client.get_database(
-            DRIVER_WORKLOAD['database']).get_collection(
-            DRIVER_WORKLOAD['collection'])
-        load_test_data(self.CONNECTION_STRING, DRIVER_WORKLOAD)
 
     def run_test(self, driver_workload):
+        self.coll = self.client.get_database(
+            [e for e in driver_workload['createEntities'] if 'database' in e][0]['database']['databaseName']).get_collection(
```
prashantmital, line 38: This is really hard to read/follow. I also think this will break if there are multiple DBs created in createEntities (not that this matters since we don't seem to currently be doing this)?
Instead, I suggest spelling out the logic explicitly so that its easy to read and its limitation are self-evident:

```
dbname = None
collname = None
for e in driver_workload['createEntities']:
    if dbname and collname:
        break
    if dbname is None and 'database' in e:
        dbname = e['database']['databaseName']
    elif collanme is None and 'collection' in e:
        collname = e['collection']['collectionName']

if not (dbname and collname):
    self.fail('Invalid scenario: executor validator test cases must provide database and collection entities')

self.coll = self.client.get_database(dbname).get_collection(collname)
```

docs/source/spec-workload-executor.rst
```diff
@@ -43,29 +43,98 @@ After accepting the inputs, the workload executor:
...
    receiving the termination signal, the workload executor:
 
    * MUST stop running driver operations and exit soon.
-   * MUST dump collected workload statistics as a JSON file named ``results.json`` in the current working directory
+   * MUST write the collected events and errors into a JSON file named
+     ``events.json`` in the current directory
+     (i.e. the directory from where the workload executor is being executed). 
+     The data written MUST be a map with the following fields:
+     
+     * ``commands``: an array of command events published during scenario
```
prashantmital, line 107: Can you modify this to say that this is an array of CommandSuccededEvent and CommandFailedEvents instead of simply calling them 'command events'?

requirements.txt
```diff
@@ -0,0 +1,8 @@
+click>=7,<8
+requests>=2,<3
+pymongo>=3.10,<4
+dnspython>=1.16,<2
+pyyaml>=5,<6
+tabulate>=0.8,<0.9
+junitparser>=1,<2
+numpy
```
prashantmital, line 8: I still don't understand fully when the `stats` CLI command is to be used but if we are to use it on Evergreen, we will also need to add `numpy` to the `install_requires` list in `setup.py`.

.evergreen/config.yml
```diff
@@ -140,30 +177,20 @@ tasks:
...
-  - name: retryWrites-toggleServerSideJS
-    tags: ["all", "retryWrites", "toggleServerSideJS"]
+          TEST_NAME: retryReads-processRestart
+  - name: retryReads-processRestart-sharded
+    cron: '@weekly'
+    tags: ["all"]
     commands:
       - func: "run test"
         vars:
-          TEST_NAME: retryWrites-toggleServerSideJS
```
prashantmital, line 166: Is there a reason you aren't running the other tests? There seem to be around 10 test files but only 2 of them have been added to the configuration here.

prashantmital (changes requested):
In order to expedite this PR, we can skip the Python code changes I have suggested in this review. I didn't want to not review the Python code since it will become harder to do so if this gets merged as-is.
There are, however, a few spec-related and general questions in my review which would need to be addressed before we can open this up to other teams.

astrolabe/runner.py
p-mongo: Changed all instances to this style except for the log downloading one where the URL is provided by the API.

astrolabe/runner.py
p-mongo: Changed!

astrolabe/cli.py
p-mongo: Removed.

astrolabe/cli.py
p-mongo: Alright, changed.

astrolabe/cli.py
p-mongo: Astrolabe aggregates the statistics so that workload executors are not obligated to do so. The statistics are uploaded as an artifact for each  test run. I personally haven't needed to see these in the terminal as part of log output which is what I think you are suggesting.

astrolabe/runner.py
p-mongo: As I understand it `self.client` is an `AtlasClient`, not a `MongoClient` hence wouldn't do the job.

p-mongo: Looking at this again I think one client instance for all iterations will suffice therefore I'll lift client instantiation to be outside of the loop.

requirements.txt
p-mongo: Done.

.evergreen/config.yml
p-mongo: Changed the configuration to run all of the tests.

astrolabe/validator.py
p-mongo: Okay I changed to this version.

docs/source/spec-workload-executor.rst
p-mongo: The structure of the data does not map to either CommandSuccededEvent or CommandFailedEvent, which is why I used the lowercase "command events" to avoid an expectation of a specific layout defined elsewhere. I am open to a different language but I think simply saying "CommandSuccededEvent or CommandFailedEvent" would not actually add clarity.

tests/retryReads-move-sharded.yaml
p-mongo: I checked this and historically the scenarios used .yaml while the validator files used .yml, so this PR continues with that split. I can change scenarios to be .yml if wanted.

p-mongo: Actually my assessment was not quite correct. The existing scenario files used .yaml. The validator didn't use a file but specified the contents inline. When I created the files I used .yml extension creating the current inconsistency. So I can either change the added .yml files to .yaml to be consistent with what this repository has historically used or I can change the .yaml files to .yml to be consistent with specs repo.

astrolabe/runner.py
prashantmital: Nice addition!

docs/source/spec-workload-executor.rst
```diff
@@ -43,29 +43,98 @@ After accepting the inputs, the workload executor:
...
+     apply URI options specified in the particular test, if any, over the
+     provided connection string.
+   
+   - Each MongoClient MUST be set up to publish `command monitoring
+     <https://github.com/mongodb/specifications/blob/master/source/command-monitoring/command-monitoring.rst>`_
+     events. The workload executor MUST record all events published
+     in the course of scenario execution.
+   
+   - Each MongoClient MUST be set up to publish `CMAP
+     <https://github.com/mongodb/specifications/blob/master/source/command-monitoring/command-monitoring.rst>`_
```
prashantmital, line 62: I think you want to link to this https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#id61 instead

docs/source/spec-workload-executor.rst
```diff
@@ -43,29 +43,98 @@ After accepting the inputs, the workload executor:
...
+     events. The workload executor MUST record all events published
+     in the course of scenario execution.
+   
+   - The ``initialData`` array in the scenario MUST be ignored by the
```
prashantmital, line 66: I do not recall why we chose to to have astrolabe load the initialData instead of letting the workload executor deal with that as well.., Might be worth filing a ticket to change this so that workload executors are once again responsible for loading initial data

docs/source/spec-workload-executor.rst
```diff
@@ -43,29 +43,98 @@ After accepting the inputs, the workload executor:
...
-     implementations should try to be as resilient as possible to these kinds of operation errors.
-   * MUST keep count of the number of operations that are run successfully (``numSuccesses``).
+     
+#. MUST use the driver's unified test runner to execute the tests in the
+   scenario, and the operations in each test, sequentially and in the order
+   in which they appear in the ``tests`` and ``operations`` arrays,
+   with the following deviations from the unified test runner specification:
+   
+   * The workload executor MUST repeat execution of the entire set of
+     specified tests and operations indefinitely, until the
```
prashantmital, line 77: 'specified tests and operations indefinitely' --> what do you mean by specified 'tests'? Isn't the whole thing that is being run a 'test'? I suggest just referring to operations and not tests.

docs/source/spec-workload-executor.rst
prashantmital: Fair enough. Let's leave it as-is and we can clarify in a new PR if language teams complain when writing their integrations.

prashantmital (changes requested):

tests/retryReads-move-sharded.yaml
jmikola: I'd vote for using `*.yml` everywhere to be consistent with our specs repo. @prashantmital: any objections?

Sorry if this comes off as bikeshedding, but I know in the specs repo we do refer to `*.yml` in scripts. I was thinking this could become an issue if https://github.com/mongodb-labs/drivers-atlas-testing/issues/100 were implemented down the line.

docs/source/spec-workload-executor.rst
jmikola: Apologies if this is already clarified elsewhere, but what is the interaction between Astrolabe and the language-specific process that executes operations? Is the intention that the test runner and its entities would be initialized once (assuming a single test) and then it would loop endlessly to execute that test until interrupted?

The Unified Test Runner on its own does not have the capability to run its operations endlessly. This sounds like a significant modification to the test runner, which wasn't mentioned in the design document ([WRITING-5844](https://jira.mongodb.org/browse/WRITING-5844)).

If the test runner isn't intended to execute `initialData`, I would suggest omitting this entirely from the document under `driverWorkload` and moving to it elsewhere. That would be similar to what we did for Atlas Data Lake (`initialData` was omitted, because ADL configs specify their own data fixtures) and save drivers from having to further modify their test runners to disable processing of the field.

That said, @p-mongo mentioned in Slack that the reason for mandating the test runner ignore `initialData` was to avoid having it execute before each repeated test.

Rather than introduce more behavioral knobs into the test runner, might I suggest a different approach of creating a new special operation type (i.e. additive change to unified test runner spec) that will take an array of operations (like `withTransaction` currently does) and execute those operations in a loop X/unlimited number of times. I'm not familiar with how the workload executor is terminated, but I imagine we can spec this out as well. For example, that special operation could be responsible for also listening for a signal to gracefully abort the loop.

astrolabe/runner.py
```diff
@@ -91,29 +93,37 @@ def is_cluster_state(self, goal_state):
...
         specification.
         """
+        
+        if no_create:
+            try:
+                self.cluster_url.get().data
+                # If --no-create was specified and the cluster exists, skip
+                # initialization. If the cluster does not exist, continue
+                # with normal creation.
+                return
```
jmikola, line 119: Is there any concern that a previous run might have left the cluster in an unhealthy state if it was terminated unexpectedly? If the cluster is found to be running, would it make sense to try and assert that its current state matches the expected initialConfiguration?

docs/source/spec-workload-executor.rst
```diff
@@ -43,29 +43,98 @@ After accepting the inputs, the workload executor:
...
+   - Any `MongoClients <https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#entity-client>`_
+     that are instantiated by the workload executor MUST use the input
+     connection string as provided to the workload executor, and MUST
+     apply URI options specified in the particular test, if any, over the
+     provided connection string.
+   
+   - Each MongoClient MUST be set up to publish `command monitoring
+     <https://github.com/mongodb/specifications/blob/master/source/command-monitoring/command-monitoring.rst>`_
+     events. The workload executor MUST record all events published
+     in the course of scenario execution.
```
jmikola, line 59: Looking at DriverWorkloadSubprocessRunner, `terminate()` issues `SIGINT` (or `CTRL_BREAK_EVENT` on Windows).

I expect most language are capable intercepting POSIX signals, but are we worried about Windows compatibility? For example, PHP's signal extension is not available on Windows and I'm not aware of any API to intercept `CTRL_BREAK_EVENT`.

Are all languages expected to run their workloads on Windows, or is Windows only supported as a possible platform for a subset of drivers?

docs/source/spec-workload-executor.rst
jmikola: Since each test will re-create entities (e.g. new MongoClients performing fresh discovery), I do think you'd need to clarify that the `operations` array is repeated in a loop rather than `tests`.

That said, see my other comment about why a special test operation may be a better way to address the needs of this project without making significant changes to the entire unified test runner.

docs/source/spec-workload-executor.rst
```diff
@@ -43,29 +43,98 @@ After accepting the inputs, the workload executor:
...
+   * The workload executor MUST keep count of the number of invocations of the scenario that
+     did not result in an error (``numSuccesses``).
 
 #. MUST set a signal handler for handling the termination signal that is sent by ``astrolabe``. The termination signal
    is used by ``astrolabe`` to communicate to the workload executor that it should stop running operations. Upon
    receiving the termination signal, the workload executor:
 
    * MUST stop running driver operations and exit soon.
-   * MUST dump collected workload statistics as a JSON file named ``results.json`` in the current working directory
+   * MUST write the collected events and errors into a JSON file named
```
jmikola, line 102: Can we clarify that exact signals that must be caught for POSIX and Windows? That would codify things in this document and avoid having folks consult the `utils.py` source.

docs/source/spec-workload-executor.rst
p-mongo: Done.

docs/source/spec-workload-executor.rst
p-mongo: I think that is an interesting idea. Adding a special operation should make it possible for the unified test runner to also insert the data, thus astrolabe wouldn't need knowledge of the unified test format at all nor would it need to integrate with Python's unified test runner.

docs/source/spec-workload-executor.rst
p-mongo: Each test consists of zero or more operations, but the unified runner executes tests, not operations. Specifying that operations must be repeated indefinitely is jumping abstraction layers and descending into unified runner's guts.

p-mongo: To clarify, existing DAT structure is scenario -> operations; with unified test format the structure becomes scenario -> tests -> operations.

tests/retryReads-move-sharded.yaml
```diff
@@ -0,0 +1,62 @@
...
+    setClusterConfiguration:
+      clusterConfiguration:
+        providerSettings:
+          providerName: AWS
+          regionName: US_EAST_1
+          instanceSizeName: M20
+      processArgs: {}
+
+uriOptions:
+  retryReads: true
```
jmikola, line 21: This is something I would have expected in the `createEntities` section of the `driverWorkload` object.

Even if Astrolabe were to process `initialData` I don't think these options would be relevant.

astrolabe/runner.py
p-mongo: I added a check for cluster configuration and if it differs from the one specified in the scenario, the cluster will be reconfigured.

docs/source/spec-workload-executor.rst
p-mongo: Ruby doesn't run on Windows at all hence I haven't run into this particular aspect of DAT.

Changing tte details of termination behavior I would say is outside the scope of this PR/spec, but I believe the following is true:

- As far as DAT is concerned, language is not required to run on a particular operating system. Any OS/es that the languages chooses to run on are fine.
- If additional changes are needed in astrolabe to support termination for new programming languages, they can in principle be made.

docs/source/spec-workload-executor.rst
p-mongo: I think that's a reasonable request but I do believe it is outside the scope of this PR, especially given that I haven't used astrolabe on windows at all and am not too familiar with the details. Could you make a new issue at https://github.com/mongodb-labs/drivers-atlas-testing/issues/new to cover this please?

tests/retryReads-move-sharded.yaml
prashantmital: I favor consistency as well. Let's use `*.yml`.

docs/source/spec-workload-executor.rst
prashantmital: @jmikola I think that is an excellent idea and probably one that would allow us to use the unified spec runner implementation without any modification in the case of drivers-atlas-testing. I do think we should do this as follow-up work though so as to not block this current work from being completed. It should be a fairly simple modification in the future for drivers to implement.

docs/source/spec-workload-executor.rst
```diff
@@ -96,8 +165,10 @@ Pseudocode Implementation
     # targetDriver is the driver to be tested.
     import { MongoClient } from "targetDriver"
 
-    # The workloadRunner function accepts a connection string and a stringified JSON blob describing the driver workload.
-    # This function will be invoked with arguments parsed from the command-line invocation of the workload executor script.
+    # The workloadRunner function accepts a connection string and a
```
prashantmital, line 168: I think this section needs to be updated to clarify what parts of the unified test runner drivers can reasonably expect to re-use and what they cannot.
For instance, you need to add comments specifying what to do with the `createEntities` field. I think the workload executor needs to take the connection string passed in by astrolabe as a string, add to it the uriOptions specified in the driverWorkload blob (read/writrePreferences etc), configure the listeners mandated by this spec (CMAP and Command) and then create the client entity. All remaining entity creation (for collection, database, session etc) can default to the normal unified format behavior.

docs/source/spec-workload-executor.rst
```diff
@@ -112,8 +183,10 @@ Pseudocode Implementation
         var num_failures = 0;
         var num_successes = 0;
 
-        # Run the workload - operations are run sequentially, repeatedly until the termination signal is received.
-        # Do not attempt to initialize the cluster with the contents of ``testData`` - astrolabe takes care of this.
+        # Run the workload - operations are run sequentially, repeatedly
+        # until the termination signal is received.
+        # Do not attempt to initialize the cluster with the contents of
+        # ``testData`` - astrolabe takes care of this.
```
prashantmital, line 189: Also add changes here for things like publishing events to a JSON file.

setup.py
```diff
@@ -27,6 +27,7 @@
     'click>=7,<8', 'requests>=2,<3',
     'pymongo>=3.10,<4', 'dnspython>=1.16,<2',
     'pyyaml>=5,<6', 'tabulate>=0.8,<0.9',
+    'numpy',
```
prashantmital, line 30: This needs to be version limited - `numpy<2`

prashantmital (changes requested):

tests/retryReads-move-sharded.yaml
p-mongo: Moved all of the uri options into createEntities and removed their handling from astrolabe entirely.

.evergreen/config.yml
```diff
@@ -335,46 +431,63 @@ buildvariants:
...
-- matrix_name: "tests-go"
-  matrix_spec:
-    driver: ["go-master"]
-    platform: ubuntu-18.04
-    runtime: go-13
-  display_name: "${driver} ${platform} ${runtime}"
-  tasks:
-    - ".all"
-  
+#- matrix_name: "tests-python"
```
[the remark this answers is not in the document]
p-mongo: Added.

astrolabe/runner.py
```diff
@@ -163,34 +176,73 @@ def run(self, persist_cluster=False, startup_time=1):
...
+                
+            if op_name == 'sleep':
+                _time.sleep(op_spec)
+                
+            if op_name == 'waitForIdle':
+                self.wait_for_idle()
+                
+            if op_name == 'restartVms':
+                rv = self.admin_client.nds.groups[self.project.id].clusters[self.cluster_name].reboot.post(api_version='private')
+                import pdb;pdb.set_trace()
```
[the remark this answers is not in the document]
p-mongo: Removed.

setup.py
p-mongo: Done.

docs/source/spec-workload-executor.rst
p-mongo: uriOptions are no longer handled by astrolabe, I am investigating removing the other requirements as well.

docs/source/spec-workload-executor.rst
jmikola: I spoke to @alcaeus about this earlier today in our 1:1 and he agreed that whether we go the route of this PR's current draft (which suggests various modifications to the test runner) or the special operation route, the unified test runner spec would likely need to be updated to outline any changes. Otherwise, we'll end up test runner modifications only being discussed in this document, which sits in a separate repository.

I got the OK to spend a day or two POC-ing a special operation for the workload executor.

p-mongo: I agree that if special operations are added, they would necessitate changes to the unified test runner. I'll push the loop POC I have so far in a moment if you'd like to review it.

docs/source/spec-workload-executor.rst
jmikola: Having the test runner repeat tests (i.e. elements in the `tests` array) would have the side effect of the runner also recreating its entities each time is starts a test. I don't see how you can avoid needing to change the test runner's internals with either approach (telling it to loop on `tests` or execute a single test and loop on `operations`); however, trying to contain all of this logic in a special operation seems like it may help avoid _most_ changes.

docs/source/spec-workload-executor.rst
p-mongo: initialData is no longer handled/touched by workload executor, OK to resolve?

docs/source/spec-workload-executor.rst
p-mongo: I see what you mean. There is now a `loop` operation defined in https://github.com/mongodb/specifications/pull/903 that allows using a single test, and the test runner no longer repeats the tests (or operations).

docs/source/spec-workload-executor.rst
```diff
@@ -37,55 +37,137 @@ Behavioral Description
...
+   executor MUST be included in the calculated (and reported) error count.
+   
+   If the unified test runner reports a failure while executing the workload,
+   the failure MUST be reported using the same format as failures handled by the
+   unified test runner, as described in the unified test runner specification
+   under the ``loop`` operation. Failures handled by the workload
+   executor MUST be included in the calculated (and reported) failure count.
+   If the driver's unified test runner is intended to handle all failures
+   internally, failures that propagate out of the unified test runner MAY
+   be treated as errors by the workload executor.
```
jmikola, line 81: The original text clarified the difference between errors and failures, but this current text seems more vague. Is there any need to differentiate between errors and failures if some implementations may not be able to?

I realize the unified test format spec says things like "consider this test to have failed" and "MUST report an error", but I think there's a lot of variation among how that is implemented in drivers. In PHP's case, where we run the unified tests as part of our larger test suite, any unexpected error (e.g. "failure" due to operation assertion failing, "error" due to unsupported operation) will bubble up and be regarded as a test failure.

To take this a step further: will most workloads even need to use result/error assertions? I was under the impression that Atlas testing mainly wanted to execute some operations in a loop and rely on features like retryable writes to minimize/avoid errors (e.g. command failures, socket exceptions). Is there a use case for wanting to assert result objects like we do for other API-centric spec tests?

docs/source/spec-workload-executor.rst
```diff
@@ -37,55 +37,137 @@ Behavioral Description
...
+   * ``numFailures``: the number of operation failures that were encountered
+     during the test. This includes failures handled by the workload executor
+     and failures handled by the unified test runner.
+   * ``numSuccesses``: the number of successful loop iterations executed
+     during the test.
+
+ .. note:: The values of ``numErrors`` and ``numFailures`` are used by ``astrolabe`` to determine the overall
+    success or failure of a driver workload execution. A non-zero value for either of these fields is construed
+    as a sign that something went wrong while executing the workload and the test is marked as a failure.
+    The workload executor's exit code is **not** used for determining success/failure and is ignored.
```
jmikola, line 161: Treating both failures and errors the same here (i.e. "test is marked as a failure") may support my earlier question about why we should differentiate them at all.

docs/source/spec-workload-executor.rst
jmikola: > configure the listeners mandated by this spec (CMAP and Command) and then create the client entity.

Do these really need to be done _before_ the client is created? I would think that command listeners could just be setup around the loop itself. As for CMAP, do most drivers allow that to be configured when constructing the MongoClient, since you might want a listener to pick up pool events from the very beginning?

docs/source/spec-workload-executor.rst
p-mongo: I believe we discussed this is as part of the unified test format PR, but to provide context in this PR: the idea behind separating errors and failures is that an error generally means the test scenario author made an error in the scenario (e.g. misspelled an operation name), whereas a failure generally means the driver under test or Atlas failed to operate as expected (e.g. a write retry failed). Thus I believe there is value in distinguishing the two, although presently the reporting mechanism for them both looks exactly the same.

p-mongo: I added a provision to the workload executor specification to set the errors or failures that are not reported by its unified test runner to the empty array.  Thus the scenarios should be written to collect both, and astrolabe should be able to assume that both would be present.

docs/source/installing-running-locally.rst
```diff
@@ -87,11 +110,11 @@ Running Atlas Planned Maintenance Tests
 
 The ``spec-tests`` command-group is used for Atlas Planned Maintenance (APM) tests. To run a single APM test, do::
 
-  $ astrolabe spec-tests run-one <path/to/test-file.yaml> -e <path/to/workload-executor> --project-name <atlasProjectName> --cluster-name-salt <randomString>
+  $ astrolabe spec-tests run-one <path/to/test-file.yml> -e <path/to/workload-executor> --project-name <atlasProjectName> --cluster-name-salt <randomString>
```
jmikola, line 113: Looking at the existing `workload-executor` implementations, it looks like they take the connection string and path to a workload file (this filename?) as their first and second arguments, respectively. Is the workload file expected to be YAML or JSON? It looks like the Python implementation uses `yaml.load`, but Node uses `JSON.parse`.

Is it reasonable to allow each language to express its preference?

I would have expected this repo to have a Makefile that converts the YAML test files to JSON, like we do in https://github.com/mongodb/specifications/blob/master/source/Makefile.

docs/source/spec-test-format.rst
```diff
@@ -31,25 +35,116 @@ A Test Scenario File has the following keys:
...
+  counts for the executed workload. The entity names for these options MUST
+  be as follows:
+  
+  - ``storeErrorsAsEntity``: ``errors``
+  - ``storeFailuresAsEntity``: ``failures``
+  - ``storeSuccessesAsEntity``: ``successes``
+  - ``storeIterationsAsEntity``: ``iterations``
+  
+  The scenario MAY use ``storeEventsAsEntities`` operation argument
+  when defining MongoClients to record CMAP events published during maintenance.
```
jmikola, line 139: Should this be changed in light of https://github.com/mongodb/specifications/pull/903/files#r574281445?

docs/source/spec-workload-executor.rst
jmikola: @prashantmital: Can you resolve this?

docs/source/spec-workload-executor.rst
jmikola: @prashantmital: I think this can be resolved now that https://github.com/mongodb/specifications/pull/903 has a `loop` operation.

docs/source/spec-workload-executor.rst
jmikola: Resolving per @prashantmital's comment that this is already documented.

docs/source/spec-workload-executor.rst
jmikola: @prashantmital: Can this be resolved? https://github.com/mongodb/specifications/pull/903 added options on the client entity definition.

docs/source/installing-running-locally.rst
p-mongo: The workload specification is mandated to be in JSON, in https://mongodb-labs.github.io/drivers-atlas-testing/spec-workload-executor.html. I agree that the test scenario handling is somewhat awkward with the scenario being defined in YAML and then a portion of it being given to the workload executor in JSON. A jira ticket could be appropriate to look at this in the future?

docs/source/spec-test-format.rst
p-mongo: Added the command monitoring bits in.

docs/source/installing-running-locally.rst
jmikola: https://jira.mongodb.org/browse/DRIVERS-1563

docs/source/spec-workload-executor.rst
```diff
@@ -37,55 +37,133 @@ Behavioral Description
...
+     option was not specified in the test scenario, the workload executor
+     MUST use ``-1`` as the number of successes.
+   
+   * ``errors``: array of documents describing the errors that occurred
+     while the workload executor was executing the operations.
+   
+   * ``failures``: array of documents describing the failures that occurred
+     while the workload executor was executing the operations.
+   
+   * ``events``: array of documents describing the CMAP events that occurred
```
jmikola, line 110: IIRC, this is no longer just CMAP events. Does this also warrant changes to the section below?

> MUST write the collected events, errors and failures into a JSON file named ``events.json`` in the current directory...

docs/source/spec-workload-executor.rst
```diff
@@ -37,55 +37,133 @@ Behavioral Description
...
+   failures were not reported by the test runner, such as because the
+   respective options were not specified in the test scenario, the workload
+   executor MUST use ``-1`` as the value for the respective counts.
+
+#. MUST write the collected events, errors and failures into a JSON file named
+   ``events.json`` in the current directory
+   (i.e. the directory from where the workload executor is being executed). 
+   The data written MUST be a map with the following fields:
+   
+   - ``events``: the collected CMAP events.
```
jmikola, line 128: The [Design Update](https://docs.google.com/document/d/1nvIY5byfFdvIYiC1XLz964U0NlvkkX_OEv1dFMronJ4/edit) was revised to refer to command monitoring events, but this doesn't talk about those events at all. Was correlation of command events (e.g. calculating durations) supposed to be added to the workload executor's responsibilities in this PR, or was that pushed to some follow-up ticket?

jmikola (commented):
Reviewed the RST files as I don't have enough context for the other components of this PR.

I have two questions on the documentation for events. If no changes are required (based on your response), you can consider me a LGTM.

docs/source/spec-workload-executor.rst
p-mongo: Yes, fixed.

docs/source/spec-workload-executor.rst
p-mongo: Correlation is performed by astrolabe, workload executor only needs to persist events as they are produced by the test runner.

astrolabe/runner.py
```diff
@@ -139,59 +144,91 @@ def run(self, persist_cluster=False, startup_time=1):
...
+                if cluster_config:
+                    LOGGER.info("Pushing cluster configuration update")
+                    self.cluster_url.patch(**cluster_config)
+
+                if process_args:
+                    LOGGER.info("Pushing process arguments update")
+                    self.cluster_url.processArgs.patch(**process_args)
+
+                # Step-4: wait until maintenance completes (cluster is IDLE).
+                self.wait_for_idle()
```
prashantmital, line 185: I think you need to sleep for a couple of seconds before calling `waitForIdle` otherwise there could be situations in which you end up seeing the IDLE server state prematurely because maintenance has not yet been kicked off by the automation agent.

astrolabe/runner.py
```diff
@@ -69,19 +72,10 @@ def cluster_url(self):
...
-                    uri_options[key] = str(value).lower()
-
-            connection_string = (prefix + "//" + self.config.database_username
-                                 + ":" + self.config.database_password + "@"
-                                 + suffix + "/?")
-            connection_string += urlencode(uri_options)
-            self.__connection_string = connection_string
+            uri = re.sub(r'://',
+                '://%s:%s@' % (self.config.database_username, self.config.database_password),
+                cluster.srvAddress)
```
prashantmital, line 77: Since the workload executors cannot add URI options to the connection string, don't we need to add them to the connection string being generated here? This was previously being done (rather simplistically) using 

```
connection_string += urlencode(uri_options)
```

but now we don't seem to be doing it at all. Can you elaborate on how workload executors are expected to respect URI options if they cannot augment the incoming connection string and if astrolabe doesn't augment the outgoing connection string?

docs/source/spec-workload-executor.rst
```diff
@@ -37,55 +37,133 @@ Behavioral Description
...
 After accepting the inputs, the workload executor:
 
-#. MUST use the input connection string to instantiate the ``MongoClient`` of the driver that is to be tested.
-   Note that the workload executor:
+#. MUST use the input connection string to `instantiate the
+   unified test runner <https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#id92>`_
+   of the driver being tested. Note that the workload executor:
 
    * MUST NOT override any of the URI options specified in the incoming connection string.
    * MUST NOT augment the incoming connection string with any additional URI options.
```
prashantmital, line 45: Is this still true? Please see my question above about how we expect to honor URI options.

prashantmital (changes requested):
You can disregard the Python style nits for this PR as that shouldn't block other drivers from implementing this spec. However, there seem to be several other questions that need to be addressed before this can be merged.

docs/source/spec-workload-executor.rst
```diff
@@ -37,55 +37,133 @@ Behavioral Description
...
-   .. note:: If ``astrolabe`` encounters an error in parsing the workload statistics dumped to ``results.json``
-      (caused, for example, by malformed JSON), ``numErrors``, ``numFailures``, and ``numSuccesses``
-      will be set to ``-1`` and the test run will be assumed to have failed.
-
-   .. note:: The choice of termination signal used by ``astrolabe`` varies by platform. ``SIGINT`` [#f1]_ is used as
-      the termination signal on Linux and OSX, while ``CTRL_BREAK_EVENT`` [#f2]_ is used on Windows.
-
-   .. note:: On Windows systems, the workload executor is invoked via Cygwin Bash.
+#. MUST parse the incoming ``driverWorkload`` document and set up
+    the driver's unified test runner to execute the provided workload.
```
jmikola, line 48: Indentation seems off here. Extra space?

jmikola (approved):
One comment about RST formatting, but content LGTM.

astrolabe/cli.py
```diff
@@ -404,6 +435,36 @@ def run_single_test(ctx, spec_test_file, workload_executor,
...
+@ATLASPROJECTNAME_OPTION
+@CLUSTERNAMESALT_OPTION
+@POLLINGTIMEOUT_OPTION
+@POLLINGFREQUENCY_OPTION
+@click.pass_context
+def get_logs_cmd(ctx, spec_test_file, org_name, project_name,
+                    cluster_name_salt, polling_timeout, polling_frequency,
+                    ):
+    """
+    Retrieves logs for the cluster.
```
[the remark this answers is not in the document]
prashantmital: The reason for suggesting this is that the docstring is what shows up as helptext if a user does `astrolabe spec-tests get-logs --help`

tests/retryWrites-toggleServerSideJS.yml
```diff
@@ -0,0 +1,64 @@
...
+driverWorkload:
+  description: "Insert"
+
+  schemaVersion: "1.0"
+
+  createEntities:
+    - client:
+        id: &client0 client0
+        uriOptions:
+          retryWrites: true
```
prashantmital, line 25: How are the uriOptions actually going to be honored? Astrolabe doesn't seem to use this field and we prohibit the workload executor from using it.

.evergreen/config.yml
```diff
@@ -78,11 +83,13 @@ functions:
           CLUSTER_NAME_SALT: ${build_id}
           ATLAS_API_USERNAME: ${atlas_key}
           ATLAS_API_PASSWORD: ${atlas_secret}
-          ATLAS_API_BASE_URL: ${atlas_url}
```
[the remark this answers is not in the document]
p-mongo: With the private endpoints the last path segment had to be removed, since I didn't have admin access I set the override in the PR. I can edit the project but this would break existing builds, therefore I suggest making this change in project config when the PR is ready to be merged.

astrolabe/runner.py
p-mongo: I reported this for failover tests in https://jira.mongodb.org/browse/PRODTRIAGE-1232 (and the users are currently meant to add the sleeps in their scenarios to compensate), are you saying other actions are also affected?

tests/retryWrites-toggleServerSideJS.yml
p-mongo: This field is read by the unified test runner when it is instantiating clients.

.evergreen/config.yml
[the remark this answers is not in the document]
prashantmital: Ok. In that case can you revert this change? We can update the `atlas_url` evergreen variable before merging this.

.evergreen/config.yml
```diff
@@ -30,6 +30,11 @@ functions:
         working_dir: astrolabe-src
         command: |
           ${PYTHON3_BINARY} -m pip install virtualenv
+    - command: subprocess.exec
```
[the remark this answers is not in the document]
prashantmital: @p-mongo what about this?

astrolabe/runner.py
prashantmital: Yes. AFAIK none of the operations on clusters are applied synchronously. Operations are put in a task queue that is consumed by the automation agent. Consequently, this could cause strange failures even for other maintenance scenarios like cluster resizing. Adding a 2-3 second wait before waiting for idle should work.

docs/source/installing-running-locally.rst
```diff
@@ -118,6 +141,17 @@ Using this flag with a given test file and static ``--cluster-name-salt`` value
...
 configuration).
 
+``astrolabe`` also provides the ``--no-create`` flag which makes it skip
+cluster initialization if the cluster already exists. This flag may be used
+to further speed up the test runs, but it can only be used for scenarios
+where the cluster configuration does not change from the initial one
+(otherwise the test would start with the wrong configuration). Using
+``--no-delete`` is recommended with ``--no-create``, otherwise each run will
+delete the cluster upon completion. **If the cluster exists but has the wrong
+configuration, ``astrolabe`` will use it as is and the results of the test
```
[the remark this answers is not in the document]
prashantmital: Can you please remve this?

docs/source/spec-test-format.rst
```diff
@@ -10,19 +10,23 @@ The YAML file format described herein is used to define platform-independent *At
...
-      document if the maintenance plan does not require modifying the Advanced Configuration Options.
-
-  * finalConfiguration (document): Description of **new** *Cluster Configuration Options* to be applied to the
-    test cluster. This document MUST contain the following keys (note that at least one of these fields MUST be
+* operations (array): List of operations to be performed, representing the
+  maintenance event. Each operation is a document containing one key which is
+  the name of the operation. The possible operations are:
+  
+  * setClusterConfiguration: set the cluster configuration to the specified
+    *Cluster Configuration Options* as defined in initialConfiguration.
```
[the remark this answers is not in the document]
prashantmital: Can you please do this? Without it, readers could assume that each operation that changes the cluster configuration would need its own `waitForIdle` operation whereas in reality, `astrolabe` automatically waits for IDLE before proceeding to the next operation.

docs/source/spec-workload-executor.rst
p-mongo: Fixed.

astrolabe/cli.py
```diff
@@ -74,29 +76,53 @@
           'the test run. Useful when a test will be run multiple times with '
           'the same cluster name salt.'))
 
+NOCREATE_FLAG = click.option(
+    '--no-create', is_flag=True, default=False,
+    help=('Do not create and configure clusters at the beginning of the run if they already exist, assume they have already been provisioned by a previous run.'))
```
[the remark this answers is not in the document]
p-mongo: Done.

astrolabe/cli.py
```diff
@@ -28,6 +29,7 @@
 from astrolabe.configuration import (
     CONFIGURATION_OPTIONS as CONFIGOPTS, TestCaseConfiguration)
 from astrolabe.utils import (
+    get_logs,
```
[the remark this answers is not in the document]
p-mongo: Which extra newline? I only see one.

astrolabe/cli.py
```diff
@@ -74,29 +76,53 @@
...
 def cli(ctx, atlas_base_url, atlas_api_username,
-        atlas_api_password, http_timeout, log_level):
+        atlas_api_password, atlas_admin_api_username, atlas_admin_api_password,
+        http_timeout, log_level):
 
     """
     Astrolabe is a command-line application for running automated driver
     tests against a MongoDB Atlas cluster undergoing maintenance.
     """
+
```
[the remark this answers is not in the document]
p-mongo: OK.

astrolabe/cli.py
```diff
@@ -16,6 +16,7 @@
 from pprint import pprint
 import unittest
 from urllib.parse import unquote_plus
+from collections import defaultdict
```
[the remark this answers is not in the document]
p-mongo: OK.

tests/retryWrites-toggleServerSideJS.yml
prashantmital: Ok, in that case we need to update the language in `docs/source/spec-workload-executor.rst`. Currently, that document states:

```
Behavioral Description
----------------------

After accepting the inputs, the workload executor:

#. MUST use the input connection string to `instantiate the
   unified test runner <https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#id92>`_
   of the driver being tested. Note that the workload executor:

   * MUST NOT override any of the URI options specified in the incoming connection string.
   * MUST NOT augment the incoming connection string with any additional URI options.
```

If we want the uriOptions to be honored, we need to update this language?

astrolabe/cli.py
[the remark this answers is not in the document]
prashantmital: Nevermind. Let's leave it as-is.

prashantmital (approved):

tests/retryWrites-toggleServerSideJS.yml
jmikola: I don't think the language you quoted conflicts. It's saying that the workload executor must pass the connection string it receives without modification to the unified test runner. AFAIK, the workload executor isn't connecting to MongoDB directly, so the only client objects using this connection string will be those created _within_ the unified test runner. IMO, this has no bearing on the unified test runner spec's instructions that a client entity's `uriOptions` should be respected.

Perhaps a clarification of this interaction would help avoid any misunderstanding.

PR #99, opened 2021-01-21
Make API version configurable when issuing a request

prashantmital:
This change allows us to specify the `apiVersion` portion of an Atlas API call at invocation (currently this is a configurable value but it can only be specified upon client initialization, not when issuing a request). The full URL generated is `<baseUrl>/<apiVersion>/<resourcePath>`.

To issue a command to the `private` API version, for example to retrieve logs, a user can now do:

```
client = AtlasClient(username=<username>, password=<password>)
response = client.nds.groups.project_id.clusters.cluster_name.get(api_version='private').data
```

### DIFF
--- a/atlasclient/client.py
+++ b/atlasclient/client.py
@@ -187,7 +187,8 @@ class AtlasClient:
             a server to the IP Whitelist, a list of documents must be sent).
         """
         method = method.upper()
-        url = self.construct_resource_url(path)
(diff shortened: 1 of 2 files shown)


prashantmital:
Failing Python tests are due to an unrelated change in PyMongo master.

p-mongo (approved):

PR #101, opened 2021-02-19
DRIVERS-828 improve diagnostics and retry log collection

p-mongo:
https://jira.mongodb.com/browse/DRIVERS-828

### DIFF
--- a/astrolabe/utils.py
+++ b/astrolabe/utils.py
@@ -279,34 +279,64 @@ def get_logs(admin_client, project, cluster_name):
         logTypes=['FTDC','MONGODB'],#,'AUTOMATION_AGENT','MONITORING_AGENT','BACKUP_AGENT'],
         sizeRequestedPerFileBytes=100000000,
     )
-    data = admin_client.groups[project.id].logCollectionJobs.post(**params).data
-    job_id = data['id']
-    
+
     # Wait 10 minutes for logs to get collected.
     # This is a guess as to how long job collection might take, we haven't
     # investigated the actual times over a sufficiently large sample size.
+    # The same timeout is used for retrying collection jobs and waiting for
+    # any single collection job - if this turns out to be problematic,
+    # we'll need to do some analysis of actual times to refine the timeouts.
     timeout = 600
-    def check():
-        data = admin_client.groups[project.id].logCollectionJobs[job_id].get().data
-        if data['status'] == 'SUCCESS':
+    local = {}
+    
+    def collect():
+        try:
+            data = admin_client.groups[project.id].logCollectionJobs.post(**params).data
+            job_id = data['id']
+            
+            def check():
+                data = admin_client.groups[project.id].logCollectionJobs[job_id].get().data
+                if data['status'] == 'SUCCESS':
+                    local['data'] = data
+                    return True
+                elif data['status'] != 'IN_PROGRESS':
+                    raise AstrolabeTestCaseError("Unexpected log collection job status: %s: %s" % (data['status'], data))
+            poll(
+                check,
+                timeout=timeout,
+                subject="log collection job '%s' for cluster '%s'" % (job_id, cluster_name),
+            )
+            
+            if 'downloadUrl' not in local['data']:
+                msg = "Log collection job did not produce a download url: %s" % data
+                del local['data']
+                raise AstrolabeTestCaseError(msg)
+                            
+            data = local['data']
+            LOGGER.info('Log download URL: %s' % data['downloadUrl'])
+            # Assume the URL uses the same host as the other API requests, and
+            # remove it so that we just have the path.
+            url = re.sub(r'\w+://[^/]+', '', data['downloadUrl'])
+            if url.startswith('/api'):
+                url = url[4:]
+            LOGGER.info('Retrieving %s' % url)
+            resp = admin_client.request('GET', url)
+            if resp.status_code != 200:
+                raise AstrolabeTestCaseError('Request to %s failed: %s' % url, resp.status_code)
+            # Note that this reads the entire response into memory, which might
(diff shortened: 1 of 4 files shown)


astrolabe/utils.py
```diff
@@ -301,15 +302,24 @@ def get_logs(admin_client, project, cluster_name):
...
         if data['status'] == 'IN_PROGRESS':
             sleep(1)
         elif data['status'] == 'SUCCESS':
+            ok = True
             break
         else:
-            raise Exception("Unexpected log collection job status %s" % data['status'])
+            raise Exception("Unexpected log collection job status: %s: %s" % (data['status'], data))
+    if not ok:
+        raise Exception("Timed out trying to collect logs from cluster %s" % cluster_name)
```
prashantmital, line 322: This should raise a PollingTimeoutError https://github.com/mongodb-labs/drivers-atlas-testing/blob/548d08035a663789eaa0c11e82002a22ad8bed2d/astrolabe/exceptions.py#L23

astrolabe/utils.py
```diff
@@ -301,15 +302,24 @@ def get_logs(admin_client, project, cluster_name):
...
+        LOGGER.info(f"Cluster %s: waiting for log collection job {job_id} for %.1f sec" % (cluster_name, timer.elapsed))
         data = admin_client.groups[project.id].logCollectionJobs[job_id].get().data
         if data['status'] == 'IN_PROGRESS':
             sleep(1)
         elif data['status'] == 'SUCCESS':
+            ok = True
             break
         else:
-            raise Exception("Unexpected log collection job status %s" % data['status'])
+            raise Exception("Unexpected log collection job status: %s: %s" % (data['status'], data))
```
prashantmital, line 320: I think we can raise AstrolabeTestCaseError here https://github.com/mongodb-labs/drivers-atlas-testing/blob/548d08035a663789eaa0c11e82002a22ad8bed2d/astrolabe/exceptions.py#L19

astrolabe/utils.py
```diff
@@ -301,15 +302,24 @@ def get_logs(admin_client, project, cluster_name):
...
     
-    while True:
-        LOGGER.debug('Poll job %s' % job_id)
+    timer = Timer()
+    timer.start()
+    ok = False
+    # Wait 10 minutes for logs to get collected - guess as to how long
+    # it might take
+    timeout = 600
+    while timer.elapsed < timeout:
```
prashantmital, line 311: We should actually subclass the PollerBase class and create an implementation for this but we can leave this be for now.

prashantmital (changes requested):
Raising bare `Exception`s is bad practice in Python so we should be raising one of the astrolabe-defined exception classes.

p-mongo:
@prashantmital  I updated the PR per your comments, extracted a poller and implemented log collection retries once the poller was extracted.

astrolabe/utils.py
```diff
@@ -300,31 +281,67 @@ def get_logs(admin_client, project, cluster_name):
...
-            break
-        else:
-            raise Exception("Unexpected log collection job status %s" % data['status'])
+    def collect():
+        try:
+            data = admin_client.groups[project.id].logCollectionJobs.post(**params).data
+            job_id = data['id']
+            
+            def check():
+                data = admin_client.groups[project.id].logCollectionJobs[job_id].get().data
```
prashantmital, line 300: `local` inside `check()` is used from the enclosing scope so we should initialize it with the `nonlocal` keyword (I know, confusing!). You can change it to:

```
def check():
    nonlocal local
    ....
```

I also suggest renaming `local` to something else to avoid confusion.

astrolabe/utils.py
```diff
@@ -300,31 +281,67 @@ def get_logs(admin_client, project, cluster_name):
...
+                if data['status'] == 'SUCCESS':
+                    local['data'] = data
+                    return True
+                elif data['status'] != 'IN_PROGRESS':
+                    raise AstrolabeTestCaseError("Unexpected log collection job status: %s: %s" % (data['status'], data))
```
prashantmital, line 305: Can you add an `else` block here? The method returns `None` when the status is `IN_PROGRESS` which allows the `poller()` method to keep polling but ideally, we should be returning a bool `False` from `check()` explicitly not implicitly.
```
elif data['status'] != 'IN_PROGRESS':
    raise AstrolabeTestCaseError("Unexpected log collection job status: %s: %s" % (data['status'], data))
else:
    # status == 'IN_PROGRESS', continue polling for logs to be ready
    return False
```

astrolabe/utils.py
```diff
@@ -300,31 +281,67 @@ def get_logs(admin_client, project, cluster_name):
...
+            if resp.status_code != 200:
+                raise AstrolabeTestCaseError('Request to %s failed: %s' % url, resp.status_code)
+            # Note that this reads the entire response into memory, which might
+            # fail for longer running workloads.
+            local['archive_content'] = resp.response.content
+            
+            return True
+        except Exception as e:
+            LOGGER.error("Error retrieving logs for '%s': %s" % (cluster_name, e))
+            # Poller will retry log collection.
```
prashantmital, line 335: Can you add an explicit `return False` here?

prashantmital (changes requested):

astrolabe/utils.py
prashantmital: More info about `nonlocal`: https://docs.python.org/3/reference/simple_stmts.html#the-nonlocal-statement

alcaeus (commented):
@prashantmital I've added your suggestions as suggested code blocks. This makes it easier for authors to apply these suggestions.

If you are happy with these changes, you can also add them to a single commit from the changed files view and have GitHub create the commit.

prashantmital (approved):

PR #102, opened 2021-03-02
Architecture overview


p-mongo:


docs/source/architecture.rst
```diff
@@ -0,0 +1,143 @@
...
+Atlas clusters. For example, a test could be to perform queries or writes
+in a loop while Atlas performs a maintenance operation, and assert that all
+queries and writes complete successfully.
+
+The project consists of the following principal components:
+
+- Astrolabe - the orchestration utility.
+- :doc:`Workload executors <spec-workload-executor>` - wrappers providing
+  a common interface to each driver's unified test runner.
+- Each driver's unified test runner.
```
prashantmital, line 14: Is this really a part of this project? I think not - it is specified and implemented separately and only consumed by the workload executor. Instead, you can rephrase the `Workload executor` component:

```
- :doc:`Workload executors <spec-workload-executor>` - wrappers around each driver's 
   unified test runner that provide a common interface for consumption by Astrolabe.
```

docs/source/architecture.rst
```diff
@@ -0,0 +1,143 @@
...
+
+Astrolabe
+---------
+
+Users invoke ``astrolabe`` to perform a test. Astrolabe interacts with Atlas
+and it performs several common tasks to keep the workload executors as small
+and simple as possible. Specifically, Astrolabe:
+
+- Creates an Atlas cluster for testing.
+- Launches a workflow executor pointed at the cluster.
```
prashantmital, line 25: Launches a workflow executor pointed at the cluster once the cluster has been provisioned and is ready for operations.*

docs/source/architecture.rst
```diff
@@ -0,0 +1,143 @@
...
+The Drivers Atlas Testing project validates the behavior of drivers against
+Atlas clusters. For example, a test could be to perform queries or writes
+in a loop while Atlas performs a maintenance operation, and assert that all
+queries and writes complete successfully.
+
+The project consists of the following principal components:
+
+- Astrolabe - the orchestration utility.
+- :doc:`Workload executors <spec-workload-executor>` - wrappers providing
+  a common interface to each driver's unified test runner.
```
prashantmital, line 13: Can we link to the unified test format here when referring to it? I think it would be best to use a link to master.

docs/source/architecture.rst
```diff
@@ -0,0 +1,143 @@
...
+
+The workload executor is invoked by Astrolabe once a test Atlas cluster is
+up and running. The workload executor:
+
+- Instantiates the driver's unified test runner.
+- Receives the workload specification, which is a unified test format test,
+  from Astrolabe, and passes it to the unified test runner.
+- Sets up a termination signal handler.
+- Invokes the unified test runner to execute the workload, which ordinarily
+  contains a ``loop`` operation.
```
prashantmital, line 64: Can you link to the `loop` operation section in the unified test format here?

docs/source/architecture.rst
```diff
@@ -0,0 +1,143 @@
...
+- When the termination signal is received from Astrolabe, instructs the
+  unified test runner to stop looping.
+- Collects the number of iterations and successful operations performed by
+  the test runner, the number of errorsr and failures, the events produced
+  by the driver while executing the workload, and writes all of this
+  to files.
+
+
+Unified Test Runners
```
prashantmital, line 73: I think with the changes I have suggested above, this section can be safely removed. IMHO we are better off without it since this (the unified format) exists independently from D-A-T (even though features have been added to it for D-A-T purposes).

docs/source/architecture.rst
```diff
@@ -0,0 +1,143 @@
...
+- Atlas maintenance operations to be performed. The operations are
+  parsed and executed by Astrolabe.
+- Driver workload to execute, which is a valid (and complete)
+  unified test format test.
+
+There are some additional requirements imposed by Drivers Atlas Testing on
+the test scenarios to simplify the implementation of both Astrolabe and
+the workload executors, namely:
+
+- There must be exactly one ``loop`` operation per test scenario.
```
prashantmital, line 104: Can you make this more specific?

- There must be exactly one ``loop`` operation per driver workload.

docs/source/index.rst
```diff
@@ -1,8 +1,21 @@
-Welcome to astrolabe's documentation!
-=====================================
+Drivers Atlas Testing
+=====================
 
-**astrolabe** is a distribution containing tools for automating Atlas operations and running Atlas Planned Maintenance
-Tests.
+The Drivers Atlas Testing project validates the behavior of drivers against
```
prashantmital, line 4: Let's not repeat text in our docs? Its a nightmare to maintain. I suggest keeping the first paragraph here and then linking to the architecture guide. The corresponding paragraph can be removed from the architecture page entirely.

docs/source/architecture.rst
```diff
@@ -0,0 +1,143 @@
+Drivers Atlas Testing Architecture
+==================================
+
+The Drivers Atlas Testing project validates the behavior of drivers against
```
prashantmital, line 4: We can remove this paragraph. See my comment on `docs/source/index.rst`

docs/source/architecture.rst
```diff
@@ -0,0 +1,143 @@
...
+
+
+Results & Artifacts
+-------------------
+
+Each test run produces the following results and artifacts:
+
+- Success or failure indication. Any errors or unified test runner failures
+  encountered during the test run fails the test run.
+- Execution log. This is useful for debugging but, because driver operations
```
prashantmital, line 116: Formatting inconsistency. Suggest changing to
`Execution log: This is useful for debugging but, because driver operations`

docs/source/architecture.rst
```diff
@@ -0,0 +1,143 @@
...
+  are performed in an infinite loop, is generally not useful for higher-level
+  analysis of the impact of Atlas maintenance on applications.
+- Server logs for all nodes in the cluster and FTDC data. Currently
```
prashantmital, line 119: Same note as above about formatting with the `:`

docs/source/architecture.rst
```diff
@@ -0,0 +1,143 @@
...
+  Drivers Atlas Testing does not use these logs but they are stored for
+  troubleshooting or investigative purposes.
+- Operation statistics: number of loop iterations performed, number of
```
prashantmital, line 122: How are the statistics distinct from the `Success or failure indication`? IIUC we derive the success or failure indiciation from the statistics so I recommend you keep this bullet point but delete the first one (i.e. delete `Success or failure indication`)

prashantmital (changes requested):
Great work putting this together @p-mongo! I have some rather granular comments so feel free to push back but I do think we should try to be as precise as possible.

docs/source/architecture.rst
p-mongo: The goal of the architecture overview document is to provide something that a newcomer to DAT might read to get an introduction to the entire system. I agree that technically unified test runner isn't part of DAT, but without mentioning it in some way I feel that the reader would be confused/mislead as to what the components are.

docs/source/architecture.rst
p-mongo: Done.

docs/source/architecture.rst
p-mongo: Done.

prashantmital:
I will port this PR to our wiki instead.

prashantmital:
Changes have been ported to the Wiki. Closing.

PR #108, opened 2021-03-09
DRIVERS-1585 Retry testFailover operations until they succeed

alcaeus:
DRIVERS-1585

I deviated from the ticket and have opted for a timeout of 90 seconds as that's the value that's used elsewhere. Happy to change it if you feel 90 seconds isn't enough.

### DIFF
--- a/astrolabe/runner.py
+++ b/astrolabe/runner.py
@@ -197,10 +197,11 @@ class AtlasTestCase:
                 while True:
                     try:
                         self.cluster_url['restartPrimaries'].post()
-                        break
                     except AtlasApiError as exc:
                         if exc.error_code != 'CLUSTER_RESTART_INVALID':
                             raise
+                    else:
+                        break
 
                     if timer.elapsed > timeout:
                         raise PollingTimeoutError("Could not test failover as cluster wasn't ready")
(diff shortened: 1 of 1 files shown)


p-mongo (approved):

astrolabe/runner.py
```diff
@@ -187,8 +187,26 @@ def run(self, persist_cluster=False, startup_time=1):
...
-                self.cluster_url['restartPrimaries'].post()
-                
+                timer = Timer()
+                timer.start()
+                timeout = 90
+
+                # DRIVERS-1585: failover may fail due to the cluster not being
+                # ready. Retry failover up to a timeout if the
+                # CLUSTER_RESTART_INVALID error is returned from the call
+                while True:
```
prashantmital, line 197: It doesn't change functionality but I would write this a bit differently to make the intent clear:

```
                while True:
                    try:
                        self.cluster_url['restartPrimaries'].post()
                    except AtlasApiError as exc:
                        if exc.error_code != 'CLUSTER_RESTART_INVALID':
                            raise
                    else:
                        break

                    if timer.elapsed > timeout:
                        raise PollingTimeoutError("Could not test failover as cluster wasn't ready")
                    else:
                        sleep(5)
```

prashantmital (approved):
LGTM though I recommend using `try...except...else` as it makes intent a lot clearer but it is a minor nit :).

I also opened #110 so that we might be able to reuse code to do things like in the future.

astrolabe/runner.py
alcaeus: Thanks for the suggestion, I wasn't aware `try...except...else` was a thing. TIL :)

PR #109, opened 2021-03-10
 DRIVERS-828 add tests for uncaptured errors and failures

p-mongo:
Test that uncaptured errors and failures fail the test run.

This PR also adds sleeps on each iteration of validator scenarios to reduce the amount of logs produced and speed up validation (since astrolabe takes some time to parse the produced json files).

### DIFF
--- a/astrolabe/utils.py
+++ b/astrolabe/utils.py
@@ -243,7 +243,9 @@ class DriverWorkloadSubprocessRunner:
 
         return self.workload_subprocess
 
-    def terminate(self):
+    def stop(self):
+        '''Stop the process, verifying it didn't already exit.'''
+        
         LOGGER.info("Stopping workload executor [PID: {}]".format(self.pid))
 
         if not self.is_windows:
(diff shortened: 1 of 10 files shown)


astrolabe/utils.py
```diff
@@ -266,7 +268,41 @@ def terminate(self):
...
             sleep(2)
+            
+        return self.read_stats()
+        
+    def terminate(self):
+        '''Ensure the process is terminated.'''
+        
+        LOGGER.info("Stopping workload executor [PID: {}]".format(self.pid))
+
+        try:
```
prashantmital, line 279: This try...except seems to be the only difference between the functions... can you explain what we can't just have one function for this?

p-mongo: Both functions combined into one per our discussion.

astrolabe/validator.py
```diff
@@ -213,10 +275,10 @@ def test_num_failures_as_errors(self):
 
         num_reported_errors = stats['numErrors']
         num_reported_failures = stats['numFailures']
-        if abs(num_reported_errors - num_reported_finds) > 1:
+        if abs(num_reported_errors - num_reported_finds/2) > 1:
```
prashantmital, line 278: Where do the `/2` come from?

astrolabe/utils.py
prashantmital: 👍🏽

astrolabe/utils.py
```diff
@@ -243,14 +243,19 @@ def spawn(self, *, workload_executor, connection_string, driver_workload,
...
-        else:
-            os.kill(self.workload_subprocess.pid, signal.CTRL_BREAK_EVENT)
-
+        
+        try:
+            if not self.is_windows:
+                os.killpg(self.workload_subprocess.pid, signal.SIGINT)
+            else:
+                os.kill(self.workload_subprocess.pid, signal.CTRL_BREAK_EVENT)
+        except ProcessLookupError as exc:
```
prashantmital, line 256: Nice! 🎉

prashantmital (commented):
Everything looks good but I don't understand what introduced all the divide-by-2 logic? Can you explain that to me?

prashantmital (approved):

PR #111, opened 2021-03-16
Add workload executor for the synchronous Java driver

jyemin:
Probably just one reviewer is sufficient, but not sure who is should be.  Feel free to choose one and remove the other.

Patch build [here](https://evergreen.mongodb.com/build/drivers_atlas_testing_tests_java__driver~java_master_platform~ubuntu_18.04_runtime~java11_patch_86b7d46cdc738755009edbc2ca08b41d85d8d2f5_6050d98c850e61097b2f157c_21_03_16_16_15_22)

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -322,10 +322,10 @@ axes:
       - id: java-master
         display_name: "Java (master)"
         variables:
-          DRIVER_DIRNAME: "java"
+          DRIVER_DIRNAME: "java/sync"
           DRIVER_REPOSITORY: "https://github.com/mongodb/mongo-java-driver"
           DRIVER_REVISION: "master"
-          ASTROLABE_EXECUTOR_STARTUP_TIME: 15
+          ASTROLABE_EXECUTOR_STARTUP_TIME: 3
       - id: dotnet-master
         display_name: "dotnet (master)"
         variables:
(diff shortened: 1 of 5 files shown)


.evergreen/config.yml
```diff
@@ -322,10 +322,10 @@ axes:
       - id: java-master
         display_name: "Java (master)"
         variables:
-          DRIVER_DIRNAME: "java"
+          DRIVER_DIRNAME: "java/sync"
```
jyemin, line 325: Eventually we may want to test the reactive streams driver, so leaving room for that here.

.evergreen/config.yml
```diff
@@ -322,10 +322,10 @@ axes:
...
           DRIVER_REPOSITORY: "https://github.com/mongodb/mongo-java-driver"
           DRIVER_REVISION: "master"
-          ASTROLABE_EXECUTOR_STARTUP_TIME: 15
+          ASTROLABE_EXECUTOR_STARTUP_TIME: 3
```
jyemin, line 328: Startup is much faster now that gradle is no longer used to run the executor

integrations/java/sync/workload-executor
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+set -o errexit
+
+workloadExecutorFile="/tmp/java-driver-workload-executor.json"
+
+rm -f "$workloadExecutorFile"
+
+# cat the workload executor JSON to a file
```
jyemin, line 9: It's easier to test the script in an IDE if the workload is passed indirectly as a file.  That's why we're not just passing the whole string as a command line parameter

integrations/java/sync/workload-executor
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+set -o errexit
+
+workloadExecutorFile="/tmp/java-driver-workload-executor.json"
```
p-mongo, line 5: Normally `mktemp` would be used instead of hardcoding the path like this.

p-mongo (approved):

PR #112, opened 2021-03-24
DRIVERS-828 remove atlas api url override


p-mongo:


alcaeus (approved):

p-mongo:
The DRIVERS-828 tests use private endpoints which require changing the "atlas" part of the original URL.

The tests were hardcoding the new path in evergreen config. I updated the config variables and removed the override.

prashantmital (approved):

PR #113, opened 2021-03-24
[V1] DRIVERS-1562 Set up periodic execution of Atlas Maintenance Tests

alcaeus:
DRIVERS-1562

This enables weekly testing of all legacy driver testing suites. Drivers should be removed from this as they are added to the evergreen config in the main development branch.

I have run a smoke-screen test with the Go driver, which initiated the test sequence correctly but timed out due while waiting for the maintenance to complete. I've chalked that up to cloud-dev being cloud-dev and assume that the tests theoretically run. I could not find configuration that completely restores the previous behaviour of not testing against cloud-dev, but that may be a feasible option to work around the cloud-dev limitations.

I've configured this patch to run all integration tests so we have a baseline for what's passing and what's not.

@jyemin should Java be removed from this?

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -141,24 +141,28 @@ tasks:
       - func: "validate executor"
   # One test-case per task.
   - name: retryReads-resizeCluster
+    cron: '@weekly'
     tags: ["all", "retryReads", "resizeCluster"]
     commands:
       - func: "run test"
         vars:
           TEST_NAME: retryReads-resizeCluster
   - name: retryReads-toggleServerSideJS
+    cron: '@weekly'
     tags: ["all", "retryReads", "toggleServerSideJS"]
     commands:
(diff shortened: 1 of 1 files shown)


alcaeus:
It looks like we're running into 20-minute polling timeouts. Should this be adjusted @prashantmital?

p-mongo (approved):

prashantmital:
Yes, I think we will need to bump the timeout... Let's use 30 mins instead of 20. 
The timeout can be increased in one of 2 ways:

1. By adding `--polling-timeout 1800` to the `astrolabe spec-tests run-one ...` invocation
2. By setting an environment variable `ATLAS_POLLING_TIMEOUT=1800` in the evergreen config.

prashantmital:
Would you like me to bump the timeouts? @alcaeus

PR #114, opened 2021-03-25
Use an evergreen variable to set the Atlas API base URL

prashantmital:
This was a regression introduced in https://github.com/mongodb-labs/drivers-atlas-testing/pull/98


prashantmital:
Dupe of #112

PR #115, opened 2021-03-29
PHPLIB-646: PHP workload executor

jmikola:
https://jira.mongodb.org/browse/PHPLIB-646

Related to mongodb/mongo-php-library#806.

Patch build: https://spruce.mongodb.com/version/60807a3dd1fe0765255b0bbd/tasks

### DIFF
--- a/integrations/php/install-driver.sh
+++ b/integrations/php/install-driver.sh
@@ -1,0 +1,86 @@
+#!/bin/bash
+
+set -o errexit  # Exit the script with error if any of the commands fail
+
+install_composer ()
+{
+    # See: https://getcomposer.org/doc/faqs/how-to-install-composer-programmatically.md
+    EXPECTED_CHECKSUM="$(php -r 'copy("https://composer.github.io/installer.sig", "php://stdout");')"
+    php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
+    ACTUAL_CHECKSUM="$(php -r "echo hash_file('sha384', 'composer-setup.php');")"
+
+    if [ "$EXPECTED_CHECKSUM" != "$ACTUAL_CHECKSUM" ]
+    then
+        >&2 echo 'ERROR: Invalid installer checksum'
+        rm composer-setup.php
+        exit 1
+    fi
+
+    php composer-setup.php --quiet
+    RESULT=$?
+    rm composer-setup.php
+
+    if [ $RESULT -ne 0 ]; then
+        exit $RESULT
+    fi
+}
+
+install_extension ()
+{
+    if [ "x${PHPC_BRANCH}" != "x" ] || [ "x${PHPC_REPO}" != "x" ]; then
+        REPO=${PHPC_REPO:-https://github.com/mongodb/mongo-php-driver}
+        BRANCH=${PHPC_BRANCH:-master}
+
+        echo "Compiling driver branch ${BRANCH} from repository ${REPO}"
+
+        BUILD_DIR="$(mktemp -d)"
+
+        git clone --depth 1 --recurse-submodules --shallow-submodules --branch ${BRANCH} ${REPO} ${BUILD_DIR}
+        pushd ${BUILD_DIR}
+
+        phpize
+        ./configure --enable-mongodb-developer-flags
+        make all -j20 > /dev/null
+        make install
+
+        popd
+    elif [ "x${PHPC_VERSION}" != "x" ]; then
+        echo "Installing driver version ${PHPC_VERSION} from PECL"
+        pecl install -f mongodb-${PHPC_VERSION}
+    else
+        echo "Installing latest driver version from PECL"
+        pecl install -f mongodb
+    fi
+
+    echo "extension=mongodb.so" > ${PHP_PATH}/lib/php.ini
+}
+
+# TODO: Remove this after https://jira.mongodb.org/browse/BUILD-13026 is resolved
+if [ -d "/opt/php/${PHP_VERSION}-64bit/bin" ]; then
+    PHP_PATH="/opt/php/${PHP_VERSION}-64bit/bin"
+else
+    # Try to find the newest version matching our constant
+    PHP_PATH=`find /opt/php/ -maxdepth 1 -type d -name "${PHP_VERSION}*-64bit" -print | sort -V -r | head -1`
+fi
+
+if [ ! -d "$PHP_PATH" ]; then
+    echo "Could not find PHP binaries for version ${PHP_VERSION}. Listing available versions..."
+    ls -1 /opt/php
+    exit 1
+fi
+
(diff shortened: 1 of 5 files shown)


integrations/php/install-driver.sh
```diff
@@ -0,0 +1,51 @@
...
+
+install_extension ()
+{
+   if [ "x${PHPC_BRANCH}" != "x" ] || [ "x${PHPC_REPO}" != "x" ]; then
+      CLONE_REPO=${PHPC_REPO:-https://github.com/mongodb/mongo-php-driver}
+      CHECKOUT_BRANCH=${PHPC_BRANCH:-master}
+
+      echo "Compiling driver branch ${CHECKOUT_BRANCH} from repository ${CLONE_REPO}"
+
+      mkdir -p /tmp/compile
```
p-mongo, line 13: I suggest either using `mktemp` or placing that under the current directory.

integrations/php/workload-executor.php
```diff
@@ -0,0 +1,45 @@
...
+
+pcntl_async_signals(true);
+pcntl_signal(SIGINT, function() {
+    Loop::allowIteration(false);
+});
+
+$tests = UnifiedTestCase::fromJSON($json);
+
+foreach ($tests as $test) {
+    $runner->run($test);
```
jmikola, line 44: Note: I believe this needs to catch exceptions and count them as either errors/failures. See [Behavior Description: #4](https://mongodb-labs.github.io/drivers-atlas-testing/spec-workload-executor.html#behavioral-description) from the workload executor docs.

integrations/php/install-driver.sh
```diff
@@ -0,0 +1,86 @@
...
+    PHP_PATH=`find /opt/php/ -maxdepth 1 -type d -name "${PHP_VERSION}*-64bit" -print | sort -V -r | head -1`
+fi
+
+if [ ! -d "$PHP_PATH" ]; then
+    echo "Could not find PHP binaries for version ${PHP_VERSION}. Listing available versions..."
+    ls -1 /opt/php
+    exit 1
+fi
+
+export PATH=${PHP_PATH}/bin:${PATH}
```
alcaeus, line 72: Nit: this could be extracted to a separate file and included using `source`. It would remove the duplicate usage and make it easier to drop once BUILD-13026 is done.

integrations/php/workload-executor.php
```diff
@@ -0,0 +1,75 @@
...
+use function MongoDB\BSON\fromJSON;
+use function MongoDB\BSON\toPHP;
+
+require_once __DIR__ . '/vendor/autoload.php';
+require_once __DIR__ . '/vendor/bin/.phpunit/phpunit/vendor/autoload.php';
+
+$uri = $argv[1];
+$json = toPHP(fromJSON($argv[2]));
+
+$entityMapObserver = function (EntityMap $entityMap) {
```
alcaeus, line 17:
```suggestion
$entityMapObserver = function (EntityMap $entityMap): void {
```

integrations/php/install-driver.sh
jmikola: I'm intentionally avoiding adding extra scripts because finding the current path to refer to them is a rabbit hole (https://stackoverflow.com/q/59895). Both `install-driver.sh` and `workload-executor` are invoked from the root directory of `drivers-atlas-testing` and changing the path within breaks `workload-executor` since it expects to write its output files in that directory. `install-driver.sh` doesn't care if we `cd`, but I'd find it irksome if both scripts ended up being inconsistent in this regard.

alcaeus: Ah, in that case let's leave this as is and focus on getting the `<major>.<minor>` links in place 👍

integrations/php/workload-executor.php
```diff
@@ -0,0 +1,111 @@
...
+    public function applyEntityMap(EntityMap $entityMap) : void
+    {
+        if ($entityMap->offsetExists('events')) {
+            $this->events = count($this->events) > 0
+                ? array_merge($this->events, (array) $entityMap['events'])
+                : (array) $entityMap['events'];
+
+            // Add placeholder CMAP events to satisfy ValidateWorkloadExecutor
+            $this->events[] = ['name' => PoolCreatedEvent, 'observedAt' => microtime(true)];
+            $this->events[] = ['name' => ConnectionCreatedEvent, 'observedAt' => microtime(true)];
```
jmikola, line 30: See: https://github.com/mongodb-labs/drivers-atlas-testing/pull/120#discussion_r613282698

integrations/php/workload-executor.php
```diff
@@ -0,0 +1,111 @@
...
+    public function applyEntityMap(EntityMap $entityMap) : void
+    {
+        if ($entityMap->offsetExists('events')) {
+            $this->events = count($this->events) > 0
+                ? array_merge($this->events, (array) $entityMap['events'])
+                : (array) $entityMap['events'];
+
+            // Add placeholder CMAP events to satisfy ValidateWorkloadExecutor
+            $this->events[] = ['name' => 'PoolCreatedEvent', 'observedAt' => microtime(true)];
+            $this->events[] = ['name' => 'ConnectionCreatedEvent', 'observedAt' => microtime(true)];
```
jmikola, line 30: See: https://github.com/mongodb-labs/drivers-atlas-testing/pull/120#discussion_r613282698

integrations/php/workload-executor.php
```diff
@@ -0,0 +1,111 @@
...
+    private $failures = [];
+    private $numSuccesses = -1;
+    private $numIterations = -1;
+
+    public function applyEntityMap(EntityMap $entityMap) : void
+    {
+        if ($entityMap->offsetExists('events')) {
+            $this->events = count($this->events) > 0
+                ? array_merge($this->events, (array) $entityMap['events'])
+                : (array) $entityMap['events'];
```
alcaeus, line 26: The count check isn't necessary, as `array_merge` with an empty array will produce the expected result. OTOH, this doesn't hurt, so I'll let you decide whether you want to make changes.

alcaeus (approved):
PHP integration LGTM besides the nit about `array_merge`.

p-mongo (approved):

integrations/php/workload-executor.php
jmikola: Ah, I wasn't sure about the performance vs. a basic assignment, but it looks like [`php_array_merge_wrapper`](https://github.com/php/php-src/blob/php-8.0.3/ext/standard/array.c#L3818) will end up doing a `ZVAL_COPY` in the common case where `$this->events` is empty since `$entityMap['events']` should be packed.

alcaeus: Ah, good to know this is actually founded in logic and not just ease to read. Thanks for checking the implementation!

PR #118, opened 2021-04-01
DRIVERS-828 log collection error should be a warning, because it will be retried


p-mongo:


matthewdale:
This PR is stale, so I'm closing it.

PR #123, opened 2021-04-21
CSHARP-3341: Comprehensive test scenario suite for planned maintenance.

DmitryLukyanov:
These changes **must not** be merged until the related .dotnet driver changes will be merged

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -408,26 +408,6 @@ axes:
         variables:
           ASYNC: "false"
           FRAMEWORK: "netcoreapp2.1"
-      - id: "dotnet-async-netcoreapp1.1"
-        display_name: dotnet-async-netcoreapp1.1
-        variables:
-          ASYNC: "true"
-          FRAMEWORK: "netcoreapp1.1"
-      - id: "dotnet-sync-netcoreapp1.1"
-        display_name: "dotnet-sync-netcoreapp1.1"
-        variables:
-          ASYNC: "false"
-          FRAMEWORK: "netcoreapp1.1"
-      - id: "dotnet-async-net452"
-        display_name: dotnet-async-net452
-        variables:
-          ASYNC: "true"
-          FRAMEWORK: "net452"
-      - id: "dotnet-sync-net452"
-        display_name: dotnet-sync-452
-        variables:
-          ASYNC: "false"
-          FRAMEWORK: "net452"
       - id: go-13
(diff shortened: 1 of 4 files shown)


.evergreen/config.yml
```diff
@@ -408,26 +408,6 @@ axes:
         variables:
           ASYNC: "false"
           FRAMEWORK: "netcoreapp2.1"
-      - id: "dotnet-async-netcoreapp1.1"
```
DmitryLukyanov, line 411: it was decided to reduce the number of used variants to reduce the atlas loading

astrolabe/utils.py
```diff
@@ -251,8 +251,9 @@ def stop(self):
 
         # Workload executors wrapped in shell scripts can report that they've
         # terminated earlier than they actually terminate on Windows.
+        # One of the reasons for this that sometimes we need to write more than N milliones log lines to the file
         if self.is_windows:
-            sleep(2)
+            sleep(60)
```
DmitryLukyanov, line 256: In most cases, it's enough to have much less timeout here, like 5-10 seconds, but very rarely even 30 seconds was not enough, so I set the safest (for now) value

astrolabe/utils.py
```diff
@@ -285,7 +286,7 @@ def get_logs(admin_client, project, cluster_name):
     params = dict(
         resourceName=rname,
         resourceType=rtype,
-        redacted=True,
+        redacted=False,
```
DmitryLukyanov, line 289: this was changed to workaround the cloud bug where this option led to unavailability to use server logs at all

astrolabe/utils.py
p-mongo: Can you please transfer this note into the source code as a comment so that developers looking at it in the future have the context for how this value was selected?

astrolabe/utils.py
p-mongo: Same with this, can you please add a note in the source code stating that redaction on 4.4 servers in Atlas produces garbled log files with a reference to the help ticket where this is determined and the cloudp ticket that would resolve the problem.

DmitryLukyanov: done

astrolabe/utils.py
DmitryLukyanov: done

p-mongo (approved):
Thank you!

PR #125, opened 2021-04-28
GODRIVER-1844 update go workload executor

iwysiu:
specification: https://mongodb-labs.github.io/drivers-atlas-testing/spec-workload-executor.html#workload-executor-specification
evergreen patch for latest version: https://spruce.mongodb.com/version/609c2ba43e8e867e6a2bbba2


integrations/go/workload_executor_test.go
```diff
@@ -0,0 +1,158 @@
...
+			}
+
+			allEvents.Failures, _ = em.BSONArray("failures")
+			allEvents.Errors, _ = em.BSONArray("errors")
+			allEvents.Events, _ = em.EventList("events")
+			// a non-nil testErr should be added to the appropriate slice
+			if testErr != nil {
+				errDoc := bson.Raw(bsoncore.NewDocumentBuilder().
+					AppendString("error", testErr.Error()).
+					AppendInt64("time", time.Now().Unix()).
```
divjotarora, line 107: In the driver's test runner code, `getSecondsSinceEpoch` returns float64 and we use `AppendDouble` to add it to the BSON documents. Should we do the same here?

integrations/go/workload_executor_test.go
```diff
@@ -0,0 +1,158 @@
...
+				switch {
+				// check for the failure substring
+				case strings.Contains(testErr.Error(), " verification failed:"):
+					allEvents.Failures = append(allEvents.Failures, errDoc)
+				default:
+					allEvents.Errors = append(allEvents.Errors, errDoc)
+				}
+			}
+
+			results.NumErrors = len(allEvents.Errors)
```
divjotarora, line 118: Per the [workload executor spec](https://mongodb-labs.github.io/drivers-atlas-testing/spec-workload-executor.html#workload-executor-specification) step 10, `numErrors` and `numFailures` should be set to -1 if the respective options were not set (I assume this means `saveErrorsAsEntity` and `saveFailuresAsEntity`).

integrations/go/workload_executor_test.go
```diff
@@ -0,0 +1,158 @@
...
+				allEvents.Events = make([]bson.Raw, 0)
+			}
+			if results.NumErrors == 0 {
+				allEvents.Errors = make([]bson.Raw, 0)
+			}
+			if results.NumFailures == 0 {
+				allEvents.Failures = make([]bson.Raw, 0)
+			}
+
+			path, _ := os.Getwd()
```
divjotarora, line 132: Why are we ignoring the error here?

integrations/go/workload_executor_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+			resultJSON, err := bson.MarshalExtJSON(&results, false, false)
```
divjotarora, line 134: Two questions:

1. Why `bson.MarshalExtJSON` rather than `json.Marshal`? I see no mention of ext JSON in the executor spec.
2. Why pass a pointer to `results` rather than the value?

divjotarora, line 134: For readability, can we add a helper function that marshals a struct to JSON and writes to a given file path?

integrations/go/workload_executor_test.go
```diff
@@ -0,0 +1,158 @@
...
+		panic(err)
+	}
+	defer func() {
+		if err := mtest.Teardown(); err != nil {
+			panic(err)
+		}
+	}()
+
+	// killAllSessions will return an auth error if it's run
+	fileReqs, runners := unified.ParseTestFile(t, workloadSpec, unified.NewOptions().SetRunKillAllSessions(false))
```
divjotarora, line 47: nit: Is the `runners` variable name a remnant from a previous version of the unified test runner changes where `ParseTestFile` returned a slice of `Runner` instances? It now returns `[]*TestCase`, so `testCases` may be a more fitting name.

integrations/go/workload_executor_test.go
```diff
@@ -0,0 +1,158 @@
...
+					c := make(chan os.Signal, 1)
+					signal.Notify(c, os.Interrupt, syscall.SIGTERM)
+
+					<-c
+					runner.EndLoop()
+				}()
+			}
+
+			testErr := runner.Run(mt)
+
```
divjotarora, line 76: nit: Remove this blank line. We generally avoid single-line blocks if possible.

integrations/go/workload_executor_test.go
```diff
@@ -0,0 +1,158 @@
...
+			em := runner.GetEntities()
```
divjotarora, line 77: nit: Give this a more descriptive name like `entityMap`.

integrations/go/workload_executor_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+			// structs for marshaling json results
+			results := struct {
```
divjotarora, line 80: All of this code looks fine, but from an organizational standpoint, I think it could be reordered for readability. I'm thinking something like:

```
var allEvents struct {
  // fields
}

// Same logic to set allEvents failures, errors, and events fields using entity map and handle testErr
// Marshal allEvents to JSON and write to file

var results struct { ... }
// use allEvents to derive fields for results
// Marshal results to JSON and write to file
```

This way, we finish working with `allEvents` before moving on to `results` rather than bouncing between them in the code.

integrations/go/workload_executor_test.go
```diff
@@ -0,0 +1,158 @@
...
+	fileReqs, runners := unified.ParseTestFile(t, workloadSpec, unified.NewOptions().SetRunKillAllSessions(false))
+
+	mtOpts := mtest.NewOptions().
+		RunOn(fileReqs...).
+		CreateClient(false)
+	mt := mtest.New(t, mtOpts)
+	defer mt.Close()
+
+	// a testfile can parse to multiple runners, though we only expect to get one per file
+	for _, runner := range runners {
```
divjotarora, line 56: How would a test file with multiple test cases work? If more than one test case had a `loop` operation, we'd be expecting multiple SIGTERM signals because each test case would block indefinitely. I don't see anything in the executor spec about this, so we should at least get that clarified in the document.

integrations/go/workload_executor_test.go
iwysiu: done

integrations/go/workload_executor_test.go
iwysiu: done

integrations/go/workload_executor_test.go
iwysiu: I'm asking in the slack channel because step 10 says that, but the pseudocode doesn't do that and the validator actually specifically checks that you don't do that.

integrations/go/workload_executor_test.go
iwysiu: un-ignored

integrations/go/workload_executor_test.go
iwysiu: gone away

iwysiu: done

integrations/go/workload_executor_test.go
```diff
@@ -0,0 +1,153 @@
...
+	for _, op := range tc.Operations {
+		if op.Name == "loop" {
+			return true
+		}
+	}
+	return false
+}
+
+// marshalStructToFile marshals the given object and writes to filePath
+func marshalStructToFile(t *testing.T, obj interface{}, filePath string) {
```
divjotarora, line 32: Add `t.Helper()` at the beginning of this function.

integrations/go/workload_executor_test.go
```diff
@@ -0,0 +1,153 @@
...
+	fileReqs, testCases := unified.ParseTestFile(t, workloadSpec, unified.NewOptions().SetRunKillAllSessions(false))
+
+	mtOpts := mtest.NewOptions().
+		RunOn(fileReqs...).
+		CreateClient(false)
+	mt := mtest.New(t, mtOpts)
+	defer mt.Close()
+
+	// a testfile can parse to multiple testCases, though we only expect to get one per file
+	for _, testCase := range testCases {
```
divjotarora, line 68: I mentioned this in my previous set of comments, but the comment is now outdated: how is the executor expected to handle multiple test cases in a single file? In your response, you linked to the part of the unified test format spec that says "Tests SHOULD NOT include multiple loop operations (nested or sequential)", but that only covers a single test case. What if the test file parses to two separate test cases, each of which has one `loop` operation?

integrations/go/workload_executor_test.go
```diff
@@ -0,0 +1,153 @@
...
+				// check for the failure substring
+				case strings.Contains(testErr.Error(), " verification failed:"):
+					allEvents.Failures = append(allEvents.Failures, errDoc)
+				default:
+					allEvents.Errors = append(allEvents.Errors, errDoc)
+				}
+			}
+
+			// make sure that empty slices marshal as slices instead of null
+			if len(allEvents.Events) == 0 {
```
divjotarora, line 117: nit: For clarity, can we switch these checks to `if <slice> == nil`? Functionally, it's the same, but it clarifies the intent of the code because we really only care about `nil` slices here.

integrations/go/workload_executor_test.go
divjotarora: Seems like the workload executor spec is incorrect and the number should actually be 0. Can you file a ticket and/or PR to update it?

iwysiu: yup, i have a pull request up

integrations/go/workload_executor_test.go
```diff
@@ -0,0 +1,156 @@
...
+				<-c
+				testCase.EndLoop()
+			}()
+		}
+
+		testErr := testCase.Run(mt)
+		entityMap := testCase.GetEntities()
+
+		// store resulting bson documents in events.json
+		allEvents := struct {
```
divjotarora, line 95: nit: For consistency with the driver code, can we declare this as `var allEvents struct {...}`. Functionally the same, but we generally use `var` syntax for declaring "empty" variables. Same for the `results` declaration below.

divjotarora (approved):
Couple of minor comments but LGTM!

integrations/go/workload_executor_test.go
```diff
@@ -0,0 +1,156 @@
...
+	fileReqs, testCases := unified.ParseTestFile(t, workloadSpec, unified.NewOptions().SetRunKillAllSessions(false))
+
+	mtOpts := mtest.NewOptions().
+		RunOn(fileReqs...).
+		CreateClient(false)
+	mt := mtest.New(t, mtOpts)
+	defer mt.Close()
+
+	// a workload must use a single test
+	if len(testCases) != 1 {
```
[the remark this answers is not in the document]
iwysiu: done

integrations/go/workload_executor_test.go
iwysiu: done

integrations/go/workload_executor_test.go
```diff
@@ -0,0 +1,156 @@
...
+		allEvents.Events, _ = entityMap.EventList("events")
+
+		// a non-nil testErr should be added to the appropriate slice
+		if testErr != nil {
+			errDoc := bson.Raw(bsoncore.NewDocumentBuilder().
+				AppendString("error", testErr.Error()).
+				AppendDouble("time", float64(time.Now().Unix())).
+				Build())
+			switch {
+			// check for the failure substring
```
kevinAlbs, line 112: This will be updated once failures within the unified test runner can be distinguished in GODRIVER-1950, right? If so, can you add the ticket number to the comment so we remember to update this check after GODRIVER-1950 is complete?

integrations/go/workload_executor_test.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+	}
+	return false
+}
+
+// marshalStructToFile marshals the given object and writes to filePath
+func marshalStructToFile(t *testing.T, obj interface{}, filePath string) {
+	t.Helper()
+	// MarshalExtJSON is used to print out bson.Raw properly
+	marshaled, err := bson.MarshalExtJSON(obj, false, false)
```
kevinAlbs, line 35: Minor, but should this instead be canonical? I don't see a reference in the workload executor specification as to whether this should be canonical or relaxed.

```suggestion
	marshaled, err := bson.MarshalExtJSON(obj, true, false)
```

integrations/go/workload_executor_test.go
```diff
@@ -0,0 +1,156 @@
...
+	"go.mongodb.org/mongo-driver/mongo/integration/unified"
+	"go.mongodb.org/mongo-driver/x/bsonx/bsoncore"
+)
+
+// hasLoop returns whether or not testCase tc contains a loop operation
+func hasLoop(tc *unified.TestCase) bool {
+	if tc == nil {
+		return false
+	}
+	for _, op := range tc.Operations {
```
kevinAlbs, line 23: No need to change. Though AFAICT it looks like the definition of [loop](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#loop) does not prohibit it appearing as a nested operation within a [withTransaction](https://github.com/mongodb/specifications/tree/master/source/transactions-convenient-api/tests#withtransaction-operation). IMO it does not seem worth adding the extra logic to check for that. I don't think there is a realistic use case for looping within a single transaction.

kevinAlbs (approved):
Apologies for the late review. LGTM with comment updates and a few minor questions. Nicely done!

integrations/go/workload_executor_test.go
p-mongo: The output should be json, not extended json.

integrations/go/workload_executor_test.go
```diff
@@ -0,0 +1,157 @@
...
+		t.Fatal(err)
+	}
+	defer func() {
+		if err := mtest.Teardown(); err != nil {
+			t.Fatal(err)
+		}
+	}()
+
+	// killAllSessions will return an auth error if it's run
+	fileReqs, testCases := unified.ParseTestFile(t, workloadSpec, unified.NewOptions().SetRunKillAllSessions(false))
```
p-mongo, line 60: While this does work, please note that https://jira.mongodb.org/browse/DRIVERS-1577 requires the unified test runner to ignore permission errors on killing sessions.

integrations/go/workload_executor_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+		allEvents.Failures, _ = entityMap.BSONArray("failures")
+		allEvents.Errors, _ = entityMap.BSONArray("errors")
+		allEvents.Events, _ = entityMap.EventList("events")
+
+		// a non-nil testErr should be added to the appropriate slice
+		if testErr != nil {
+			errDoc := bson.Raw(bsoncore.NewDocumentBuilder().
+				AppendString("error", testErr.Error()).
+				AppendDouble("time", float64(time.Now().Unix())).
```
p-mongo, line 109: My reading of https://golang.org/pkg/time/ suggests that this call truncates milliseconds even though it outputs a float. Therefore I imagine it could be possible for error times to be non-sequential (if the unified test runner correctly records times with millisecond precision) or, if the test runner also truncates precision, the times may be difficult to work with as many of them would be identical for events that didn't happen at the same time.

p-mongo (approved):

integrations/go/workload_executor_test.go
iwysiu: Yeah, the relaxed extended json ends up being identical to json for results.json, and for events.json actually produces human readable events documents.

integrations/go/workload_executor_test.go
iwysiu: removed. This was written before the go ticket for that was done.

integrations/go/workload_executor_test.go
iwysiu: Good catch, I think I misunderstood the spec the first time I read it. Is this more what you were thinking?

p-mongo: The new version looks good to me, thanks!

p-mongo (approved):

PR #131, opened 2021-11-09
PYTHON-2512 Update Astrolabe's Workload Executor to use the unified test runner

juliusgeo:


juliusgeo:
It's not quite ready for review as I still need to verify it is giving the correct results, however it looks like it works so far.

juliusgeo:
@ShaneHarvey Should be ready for review.

integrations/python/pymongo/workload-executor.py
```diff
@@ -35,92 +22,66 @@ def interrupt_handler(signum, frame):
...
+    runner.TEST_SPEC = test_workload
+    UnifiedSpecTestMixinV1.TEST_SPEC = test_workload
+    runner.setUpClass()
+    runner.setUp()
+    # should be removed in final version, but useful for testing right now
+    for op in test_workload["tests"][0]["operations"]:
+        if op.get("name") == "loop":
+            op["arguments"]["numIterations"] = 10
+    try:
+        runner.run_scenario(test_workload["tests"][0], uri=mongodb_uri)
```
ShaneHarvey, line 47: We should assert that test_workload["tests"] only has one test in it to make sure we're not silently ignoring extra tests.

integrations/python/pymongo/workload-executor.py
```diff
@@ -35,92 +22,66 @@ def interrupt_handler(signum, frame):
...
-    exit(0 or num_errors or num_failures)
+    runner = UnifiedSpecTestMixinV1()
+    runner.TEST_SPEC = test_workload
+    UnifiedSpecTestMixinV1.TEST_SPEC = test_workload
+    runner.setUpClass()
+    runner.setUp()
+    # should be removed in final version, but useful for testing right now
+    for op in test_workload["tests"][0]["operations"]:
+        if op.get("name") == "loop":
+            op["arguments"]["numIterations"] = 10
```
ShaneHarvey, line 45: Remove this? Alternatively you could control this behavior via a command line argument to the workload-executor.py script.

integrations/python/pymongo/run-for-all.sh
```diff
@@ -0,0 +1,9 @@
+for file in ../../../tests/*
+do
+   if [[ $file == *.yml ]]
+   then
+      echo $file
+      python3 workload-executor.py "user:password@localhost" "$file"
```
ShaneHarvey, line 6: "user:password@localhost" is not a valid connection string because it's missing "mongodb://". How does this work?

integrations/python/pymongo/run-for-all.sh
```diff
@@ -0,0 +1,9 @@
+for file in ../../../tests/*
```
ShaneHarvey, line 1: Make this the first line:
```
#!/bin/bash -ex
```

Also please add a comment that explains what this file is supposed to do and why it's useful.

integrations/python/pymongo/workload-executor.py
```diff
@@ -35,92 +22,66 @@ def interrupt_handler(signum, frame):
...
-        assert result == expected_result
-
-
-def connect(mongodb_uri):
-    if WIN32 and mongodb_uri.startswith("mongodb+srv://"):
-        # TODO: remove this once BUILD-10841 is done.
-        import certifi
-        return MongoClient(mongodb_uri, tlsCAFile=certifi.where())
-    return MongoClient(mongodb_uri)
+def filter_failures_errors(entity_map):
```
ShaneHarvey, line 25: This method is unused.

ShaneHarvey (changes requested):

integrations/python/pymongo/workload-executor.py
juliusgeo: Done.

integrations/python/pymongo/run-for-all.sh
juliusgeo: I am honestly not sure how it works. Could it mean that my routing in pymongo is incorrect?

integrations/python/pymongo/run-for-all.sh
juliusgeo: Done.

integrations/python/pymongo/workload-executor.py
```diff
@@ -35,92 +22,66 @@ def interrupt_handler(signum, frame):
...
+        runner.run_scenario(test_workload["tests"][0], uri=mongodb_uri)
+    except Exception as exc:
+        target = "failures" if isinstance(exc, AssertionError) else "errors"
+        if target not in runner.entity_map:
+            runner.entity_map[target] = []
+        runner.entity_map[target].append({
+            "error": str(exc),
+            "time": time.time(),
+            "type": type(exc)
+        })
```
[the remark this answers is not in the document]
juliusgeo: Actually I don't think that that test proves my point. However, there is a section of the pseudocode implementation that I believe is what inspired my approach 
``` 
try {
        runner.executeScenario();
    } catch (propagatedError) {
        /* If the test runner propagates an error or failure (e.g. it is not
         * captured by the loop or occurs outside of the loop), it MUST be
         * reported by the workload executor. */
         errors.push({
           error: propagatedError.message,
           time: Date.now() / 1000
         });
    }
```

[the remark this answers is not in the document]
juliusgeo: It also states this about error propagation:
"Note that is possible for some or all of these arrays to be empty if the corresponding data was not reported by the unified test runner and the test runner did not propagate an error or failure (which would then be reported by the workload executor)."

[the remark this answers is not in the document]
juliusgeo: Also I tried it out without the error catching code and the spec says we MUST write out results.json and events.json, but there are assertion errors not being caught by pymongo that cause the program to exit without writing statistics.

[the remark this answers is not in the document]
juliusgeo: An example of a test that causes the workload executor to throw an error (rather than it being caught in the loop) is /tests/validator-numFailures-not-captured.yml

.evergreen/config.yml
```diff
@@ -352,8 +352,8 @@ axes:
         display_name: "PyMongo (master)"
         variables:
           DRIVER_DIRNAME: "python/pymongo"
-          DRIVER_REPOSITORY: "https://github.com/mongodb/mongo-python-driver"
-          DRIVER_REVISION: "master"
+          DRIVER_REPOSITORY: "https://github.com/juliusgeo/mongo-python-driver"
+          DRIVER_REVISION: "PYTHON-3014"
```
ShaneHarvey, line 356: After merging PYTHON-3014, let's revert this.

ShaneHarvey (changes requested):

.evergreen/config.yml
juliusgeo: Done.

integrations/python/pymongo/workload-executor.py
```diff
@@ -1,31 +1,22 @@
...
 
-from pymongo import MongoClient
-from pymongo.cursor import Cursor
-from pymongo.command_cursor import CommandCursor
-
-
-IS_INTERRUPTED = False
+from collections import defaultdict
+import pymongo
+test_path = os.path.dirname(os.path.dirname(inspect.getfile(pymongo)))
```
ShaneHarvey, line 12: A comment that explains what and why this is required would be helpful.

ShaneHarvey (approved):
LGTM once the tests (mostly) pass. As discussed please open a new issue to investigate the flakey tests.

integrations/python/pymongo/workload-executor.py
```diff
@@ -35,92 +26,42 @@ def interrupt_handler(signum, frame):
...
-    sentinel = os.path.join(os.path.abspath(os.curdir), 'results.json')
-    print("Writing statistics to sentinel file {!r}".format(sentinel))
-    with open('results.json', 'w') as fr:
-        json.dump(stats, fr)
-    exit(0 or num_errors or num_failures)
+    runner = UnifiedSpecTestMixinV1()
+    runner.TEST_SPEC = test_workload
+    UnifiedSpecTestMixinV1.TEST_SPEC = test_workload
+    runner.setUp()
+    runner.client = pymongo.MongoClient(mongodb_uri)
```
juliusgeo, line 34: This is necessary because there does not seem to exist a localhost:27017 instance of mongodb, so this instead overrides the client with the mongodb_uri that is where mongodb is actually located.

ShaneHarvey: Please explain this in a comment as well.

juliusgeo: Done.

integrations/python/pymongo/workload-executor.py
juliusgeo: Done.

ShaneHarvey (approved):
LGTM

PR #134, opened 2022-05-06
CLOUDP-100718 Improve Astrolabe Testing

erictsai6:
Improvements made:
* Searching by organization name is inefficient as you have to page (the code doesn't look like it's paging either) through thousands of organizations before finding a string match.  There's a high chance of naming collision too since name is not unique. 
    * This will require all env variables to be updated to org id rather than org name
* Creates a new project per AtlasTestCase with the suffix being the epoch timestamp in seconds. Ex: [ERICATEST-1651857784](https://cloud-qa.mongodb.com/v2/627559788a1e082d1ad4bbfc#)
* Cleans up any old projects in the organizations that have an epoch timestamp that exceeds 12 hours from current time.

### DIFF
--- a/astrolabe/commands.py
+++ b/astrolabe/commands.py
@@ -22,17 +22,11 @@ from atlasclient import AtlasApiError
 LOGGER = logging.getLogger(__name__)
 
 
-def get_one_organization_by_name(*, client, organization_name):
-    """Get the ID of the organization by the given name. Raises
-    AtlasApiError if no organization exists by the given name."""
-    all_orgs = client.orgs.get().data
-    for org in all_orgs.results:
-        if org.name == organization_name:
-            LOGGER.debug("Organization details: {}".format(org))
-            return org
-
-    raise AtlasApiError('Organization {!r} not found.'.format(
-        organization_name))
+def get_organization_by_id(*, client, org_id):
+    """Get the organization by the given id `org_id`."""
+    org = client.orgs[org_id].get().data 
+    LOGGER.debug("Organization details: {}".format(org))   
+    return org    
 
 
 def ensure_project(*, client, project_name, organization_id):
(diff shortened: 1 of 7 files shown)


astrolabe/runner.py
```diff
@@ -383,6 +383,28 @@ def __init__(self, *, client, admin_client, test_locator_token, configuration, x
...
+        # Set up Atlas for tests.
+        # Step-1: ensure validity of the organization.
+        # Note: organizations can only be created by via the web UI.
+        org_id = self.config.organization_id
+        LOGGER.info("Verifying organization id: {!r}".format(org_id))
+        org = get_organization_by_id(
+            client=self.client, org_id=org_id)
+        LOGGER.info("Successfully verified organization {!r}".format(org.name))
+
+        # Step-2: purge old projects from organization.
```
p-mongo, line 398: I think this ultimately would be better off as a separate build step after the tests complete.

p-mongo (approved):

astrolabe/commands.py
```diff
@@ -22,6 +22,16 @@
...
+    """Get the organization by the given id `org_id`."""
+    org = client.orgs[org_id].get().data 
+    LOGGER.debug("Organization details: {}".format(org))   
+    return org    
+
+
+# TODO(etsai) - this method is inefficient, DO NOT USE.  
+# We have thousands of organizations that we'd have to page through 
+# and there's the risk of naming collision
 def get_one_organization_by_name(*, client, organization_name):
```
matthewdale, line 35: Is there any reason to keep this function around? If we shouldn't use it, we should remove it.

matthewdale (approved):
This is awesome 👍

astrolabe/commands.py
erictsai6: will remove

astrolabe/runner.py
erictsai6: I can create it as a separate build step but I think it's more beneficial before the test runs.  Does that work with you @p-mongo ?

erictsai6:
Removed the get_organization_by_name and put the project clean into its own method for readability 

```
INFO:astrolabe.runner:Verifying organization id: '6273ccbfeed3323ccea42b2f'
INFO:astrolabe.runner:Successfully verified organization 'abc'
INFO:astrolabe.runner:Successfully deleted project 'ericaTest-111', id: '627a9c992004517c3a4fe28c'
INFO:astrolabe.runner:Loading spec test from file '/Users/eric.tsai/drivers-atlas-testing/tests/retryReads-move.yml'
INFO:astrolabe.runner:Verifying project 'ericaTest-1652202689'
INFO:astrolabe.runner:Successfully verified project 'ericaTest-1652202689'
INFO:astrolabe.runner:Verifying user 'atlasuser'
INFO:astrolabe.runner:Successfully verified user 'atlasuser'
INFO:astrolabe.runner:Enabling access from anywhere on project 'ericaTest-1652202689'
INFO:astrolabe.runner:Successfully enabled access from anywhere on project 'ericaTest-1652202689'
INFO:astrolabe.runner:Astrolabe Test Plan
===============  ====================
Test name        Atlas cluster name
===============  ====================
retryReads_move  173cf02553
===============  ====================
```

PR #135, opened 2022-05-31
RUST-1300 Add Rust integration

isabelatkinson:


astrolabe/validator.py
```diff
@@ -98,7 +98,7 @@ def run_test_expecting_error(self, driver_workload):
         try:
             try:
                 stats = subprocess.stop()
-            except PrematureExitError:
+            except (PrematureExitError, PermissionError):
```
isabelatkinson, line 101: The expected `PrematureExitError` here was showing up as a `PermissionError` for me when the validator attempted to kill an already-terminated process.

p-mongo: Could you add this as a comment in the code please so that whoever looks at this section in the future is aware of this case?

p-mongo (approved):

.evergreen/config.yml
```diff
@@ -574,3 +583,11 @@ buildvariants:
   display_name: "${driver} ${platform} ${runtime}"
   tasks:
     - ".all"
+- matrix_name: "tests-rust"
+  matrix_spec:
+    driver: ["rust-main"]
+    platform: ubuntu-18.04
```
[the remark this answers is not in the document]
kmahar: just linux seems ok to me for now

integrations/rust/install-driver.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+set -o errexit
+set -o xtrace
+
+rm -rf ~/.rustup
+curl https://sh.rustup.rs -sSf | sh -s -- -y --no-modify-path
```
kmahar, line 7: should we be installing rustup and cargo in the project directory rather than home directory, similar to what we switched to for the driver?

kmahar: I think it's not strictly necessary in the way that it was for the driver -- it came up there because the documentation task does some chmod-ing to simulate a read-only file system, which can break future builds that use the same cargo install. 
that said, per Evergreen team it's considered best practice to not write to the home directory on the hosts since they do get reused:

> Best practice for evergreen is to stick all files in the task working directory rather than writing to the home directory

patrickfreed (approved):
LGTM!

And just to confirm: no evergreen tasks were supposed to be kicked off as part of this PR right?

isabelatkinson:
I'm not sure why the patches aren't running; I’ve been doing them manually and will run a final one before merging.

abr-egn (approved):
LGTM!

PR #137, opened 2022-07-11
 DRIVERS-2365 Append random number suffix to make project names more unique

p-mongo:


astrolabe/runner.py
```diff
@@ -395,8 +397,9 @@ def __init__(self, *, client, admin_client, test_locator_token, configuration, x
             client=self.client, org_id=org_id)
         LOGGER.info("Successfully verified organization {!r}".format(org.name))
 
-        # Step-2: clean old projects from organization.
-        self.clean_old_projects(org.id)                             
+        # Step-2: clean old projects with same name base from organization.
+        if True or not no_create:
+            self.clean_old_projects(org.id)
```
matthewdale, line 402: Is forcing this condition to be `True` intentional? If not, remove the `True or`.

matthewdale (approved):
Looks good 👍

PR #143, opened 2022-08-23
Remove errors/failures 'not captured' workload executor validation tests.

matthewdale:
The `test_num_errors_not_captured` and `test_num_failures_not_captured` tests assert that the workload executor will still succeed and write a `results.json` file when `storeErrorsAsEntity` and `storeFailuresAsEntity` are both excluded from the spec test file. However, the [spec test format](https://github.com/mongodb-labs/drivers-atlas-testing/blob/0b1029caa2e20479cef93757249588e4ce1ce6b0/docs/source/spec-test-format.rst) explicitly requires including the `storeErrorsAsEntity` argument:
> The scenario MUST use storeErrorsAsEntity, storeSuccessesAsEntity, and storeIterationsAsEntity operation arguments to allow the workload executor to retrieve errors, failures, and operation and iteration counts for the executed workload.

Testing that the workload executors work without the `storeErrorsAsEntity` argument seems unnecessary.

Currently the Java, PHP, Rust, and .NET workload executors fail those two validator tests. When they encounter errors or failures, the workload executor exits with a non-zero exit code and doesn't write a `results.json` file, resulting in a test failure. That actually seems like preferable behavior compared to the passing workload executors, which effectively ignore execution errors/failures.


matthewdale:
Some remaining validator tests for the Rust and PHP driver still fail reliably, which may need to be investigated separately.

p-mongo:
I think the relevant behavior is specified in https://mongodb-labs.github.io/drivers-atlas-testing/spec-workload-executor.html section 5:

> If the unified test runner raises an error while executing the workload, the error MUST be reported using the same format as errors handled by the unified test runner, as described in the unified test runner specification under the loop operation. The error MUST be appended to the errors array.

If the files in question cause the test runner to raise an error or otherwise fail, what should happen is the workload executor is supposed to handle those errors and exit with the proper files written. If that is not happening I would say the workload executor is not compliant with the requirements.

PR #146, opened 2022-09-07
DRIVERS-2420 Use a keystore to provide TLS certs to the Java driver.

matthewdale:
[DRIVERS-2420](https://jira.mongodb.org/browse/DRIVERS-2420)

The Java driver does not support the `tlsCertificateKeyFile` and `tlsCAFile` connection string parameters (see [JAVA-3066](https://jira.mongodb.org/browse/JAVA-3066)) and instead uses a "key store" to provide TLS certs and CAs to the JVM. Update the Java workload executor script to check for the `tlsCertificateKeyFile` and `tlsCAFile` connection string parameters and, if present, add them to key stores and point the JVM at those key stores.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -733,11 +733,7 @@ buildvariants:
     runtime: ["java17"]
   display_name: "${driver} ${platform} ${runtime}"
   tasks:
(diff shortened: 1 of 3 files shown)


jyemin (approved):
It all looks good, so if the patch succeeds then LGTM

p-mongo (approved):

PR #147, opened 2022-09-07
Update .NET build SDK version to 6.0

matthewdale:
The .NET driver started failing recently (e.g. [here](https://evergreen.mongodb.com/lobster/evergreen/task/drivers_atlas_testing_tests_dotnet_linux__driver~dotnet_master_platform~ubuntu_18.04_runtime~dotnet_async_netcoreapp2.1_kind_reads_baseline_631825530305b9103d6b7d46_22_09_07_05_00_03/0/task#bookmarks=0%2C206&l=1)) with an error message like:
```
error NETSDK1045: The current .NET SDK does not support targeting .NET 6.0.  Either target .NET 5.0 or lower, or use a version of the .NET SDK that supports .NET 6.0. [/mongo-csharp-driver/tests/MongoDB.Bson.Tests/MongoDB.Bson.Tests.csproj]
```
It seems to be related to https://github.com/mongodb/mongo-csharp-driver/pull/873, which updates the target framework for the `MongoDB.Bson.Tests.csproj` file.

Update the .NET build script to download v6.0 instead of v5.0

### DIFF
--- a/integrations/dotnet/install-driver.sh
+++ b/integrations/dotnet/install-driver.sh
@@ -26,9 +26,9 @@ if [[ "$OS" =~ Windows|windows ]]; then
         -o integrations/dotnet/dotnet-install.ps1 \
         https://dot.net/v1/dotnet-install.ps1
 
-    # Install the v5.0 SDK to build the driver and the v2.1 SDK to run the published workload
+    # Install the v6.0 SDK to build the driver and the v2.1 SDK to run the published workload
     # executor binaries.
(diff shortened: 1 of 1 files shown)


p-mongo (approved):

JamesKovacs (approved):
While bumping the .NET version from 5 to 6 fixes the immediate problem, we still have the longterm pain of having to adjust Astrolabe every time the driver has to bump versions. It would be much more maintainable to add `AstrolabePublish` and `AstrolabeRun` tasks to the driver's `build.cake` script. I've created https://jira.mongodb.org/browse/CSHARP-4315 to tackle this work.

In the short term, I'll LGTM this change to get Astrolabe back up and running successfully.

PR #150, opened 2022-09-12
DRIVERS-2263 Add a Kind test case for server version upgrade.

matthewdale:
[DRIVERS-2263](https://jira.mongodb.org/browse/DRIVERS-2263)

Add a Kind test case that upgrades the server version from 5.0.8 to 5.0.12. Assert that reads and writes (with retry enabled) can continue successfully during the upgrade.

Also add some delay to the read and write loops so that the number of events collected doesn't become unreasonably large, especially for the local Kind cluster where the round trip time is very short.

### DIFF
--- a/tests/kubernetes/kind/upgrade.yml
+++ b/tests/kubernetes/kind/upgrade.yml
@@ -1,0 +1,12 @@
+# Upgrade the MongoDB server version from 5.0.8 to 5.0.12. That will cause a rolling deploy where
+# each pod is terminated and recreated with the new server version. This process typically takes
+# between 5 and 10 minutes for a 3-node replica set.
+# This test case is used to assert that reads and writes can continue when a rolling deploy of the
+# MongoDB replica set happens.
+operations:
+  - kubectl: [--namespace, default, patch, MongoDBCommunity, mongodb, --type=merge, -p, '{"spec":{"version":"5.0.12"}}']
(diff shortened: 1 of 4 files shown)


tests/kubernetes/kind/upgrade.yml
```diff
@@ -0,0 +1,12 @@
+# Upgrade the MongoDB server version from 5.0.8 to 5.0.12. That will cause a rolling deploy where
```
p-mongo, line 1: Where is the 5.0.8 version specified?

p-mongo (approved):

tests/kubernetes/kind/upgrade.yml
matthewdale: 5.0.8 is specified in the `MongoDBCommunity` resource in [mongo.yml](https://github.com/mongodb-labs/drivers-atlas-testing/blob/8b3f72b040045399ed4cd9e35d440f8eea314bf1/kubernetes/kind/mongodb.yml#L21).

matthewdale: Good idea, will add comments!

matthewdale: Added comments in https://github.com/mongodb-labs/drivers-atlas-testing/pull/151.

PR #157, opened 2022-10-03
DRIVERS-2418 Fix JUnit output to only indicate failure when failure isn't expected.

matthewdale:
[DRIVERS-2418](https://jira.mongodb.org/browse/DRIVERS-2418)

For some reason, the JUnit test results for the "expect failure" tasks were not uploaded before https://github.com/mongodb-labs/drivers-atlas-testing/pull/155 was merged, so the tasks passed despite the JUnit test output being marked failure. Update the JUnit test results to only indicate failure if "expect failure" and the actual failure do not match.

### DIFF
--- a/astrolabe/atlas_runner.py
+++ b/astrolabe/atlas_runner.py
@@ -280,6 +280,8 @@ class AtlasTestCase:
             # Step-5: interrupt driver workload and capture streams
             stats = self.workload_runner.stop()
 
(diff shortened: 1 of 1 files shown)


matthewdale:
I've independently tested that these changes resolve some existing failures. As there is currently no secondary spec owner for Astrolabe and these changes are low risk, I'm merging this without review.

PR #158, opened 2022-10-05
DRIVERS-2460 Use python3 toolchain in /opt/python/ instead of /opt/mongodbtoolchain/

matthewdale:
[DRIVERS-2460](https://jira.mongodb.org/browse/DRIVERS-2460)

The version of Python that ships with the v4 "mongodbtoolchain" is now Python 3.10. Use a Python toolchain in `/opt/python` instead of `/opt/mongodbtoolchain` so we use a consistent version of Python.

Also remove the unnecessary `PYTHON_BINARY` variables from the Java and Go runtime axes.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -683,7 +683,7 @@ axes:
         run_on: ubuntu1804-drivers-atlas-testing
         batchtime: 10080  # 7 days
         variables:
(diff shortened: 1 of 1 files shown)


matthewdale:
Confirmed these changes fix the Astrolabe run failures. Merging without review.

PR #160, opened 2022-12-20
Catch and retry HTTP timeouts that happen while polling for cluster state.

matthewdale:
Sometimes the `astrolabe` CLI will exit when polling for Atlas cluster status due to an HTTP call timeout. An example exception output is:
```
atlasclient.exceptions.AtlasClientError: HTTPSConnectionPool(host='cloud-qa.mongodb.com', port=443): Read timed out. (read timeout=30.0) (GET https://cloud-qa.mongodb.com/api/atlas/v1.0/groups/639e9f7f25930c386c4e5ad3/clusters/4ba448af53)
```

Those HTTP call timeouts seem to be intermittent, so we should retry the call again instead of allowing the CLI to exit.

### DIFF
--- a/astrolabe/atlas_runner.py
+++ b/astrolabe/atlas_runner.py
@@ -338,7 +338,14 @@ class AtlasTestCase:
         wanted_state = 'idle'
         last_notified = 0
         while timer.elapsed < timeout:
-            cluster_info = self.cluster_url.get().data
+            sleep(1.0 / self.config.polling_frequency)
+
+            try:
+                cluster_info = self.cluster_url.get().data
+            except AtlasClientError as e:
+                LOGGER.error(f"Error getting cluster status: {e}")
+                continue
+
             actual_state = cluster_info.stateName.lower()
             if actual_state == wanted_state:
                 ok = True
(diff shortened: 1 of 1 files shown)


astrolabe/atlas_runner.py
```diff
@@ -338,7 +338,14 @@ def wait_for_idle(self):
         wanted_state = 'idle'
         last_notified = 0
         while timer.elapsed < timeout:
-            cluster_info = self.cluster_url.get().data
+            sleep(1.0 / self.config.polling_frequency)
```
ShaneHarvey, line 341: I was confused by this until I realized polling_frequency is actually a frequency and not a timespan like heartbeatFrequencyMS.

ShaneHarvey (approved):

astrolabe/atlas_runner.py
matthewdale: Yeah, interval is typically more intuitive than frequency. Technically `heartbeatFrequencyMS` is a misnomer and should be called `heartbeatIntervalMS` because it defines an interval.

matthewdale:
Verified this change works in the "happy path". My testing didn't encounter any HTTP timeout errors calling the cluster status API, so I haven't verified it actually fixes the problem. I'm going to merge it and see if it fixes the problem on the next scheduled run.

PR #161, opened 2023-02-04
Switched Astrolabe to run on .NET 6.

JamesKovacs:
.NET Core 2.x is EOL and Microsoft removed the binaries from the download location. Switching to .NET 6 since it is an LTS release.

WARNING: I was unsuccessful in running `astrolabe` locally. So these changes are currently untested. If there are any issues, please let me know and I can investigate.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -716,16 +716,16 @@ axes:
         display_name: Java 17
         variables:
           JAVA_HOME: "/opt/java/jdk17"
-      - id: dotnet-async-netcoreapp2.1
-        display_name: dotnet-async-netcoreapp2.1
+      - id: dotnet-async-net6
+        display_name: dotnet-async-net6
         variables:
           ASYNC: "true"
-          FRAMEWORK: "netcoreapp2.1"
-      - id: "dotnet-sync-netcoreapp2.1"
-        display_name: dotnet-sync-netcoreapp2.1
+          FRAMEWORK: "net6.0"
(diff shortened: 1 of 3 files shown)


matthewdale (commented):
Tested in a patch build. The workload executor fails to launch with the following error:
```
You must install or update .NET to run this application.
App: /astrolabe-src/mongo-csharp-driver/tests/AstrolabeWorkloadExecutor/bin/Debug/netcoreapp2.1/publish/workload-executor.dll
Architecture: x64
Framework: 'Microsoft.NETCore.App', version '2.1.0' (x64)
.NET location: /astrolabe-src/.dotnet/
```
Is that possibly caused by the target frameworks including `netcoreapp2.1` in [this configuration](https://github.com/mongodb/mongo-csharp-driver/blob/5d1353dc000613c46184636c8a30c86c46970905/tests/BuildProps/Tests.Build.props#L9)?

JamesKovacs:
Totally forgot about this PR. I found the reference to .NET Core 2.1 in the `.evergreen/config.yml`. I switched it to only run on .NET 6 to permit upgrading Astrolabe to Ubuntu 22.04.

matthewdale (approved):
Hah, better late than never! Looks good 👍

JamesKovacs:
Note that this PR is failing in Astrolabe due to test failures apparently from DRIVERS-1652 not being implemented in C# yet (CSHARP-3826). Merging this PR so that we can work on fixing the Astrolabe test runner in C#.

PR #170, opened 2023-08-15
GODRIVER-2946 Remove call to mtest.T.Close in Go integration.

matthewdale:
The Go driver recently removed the `mtest.T.Close` function, which is now called automatically (see https://github.com/mongodb/mongo-go-driver/pull/1347). Stop calling it in the Go integration.


blink1073 (approved):
LGTM!

PR #172, opened 2023-08-23
test(DRIVERS-2700): remove failure also from evg run

durran:
failure.yml was removed but still expected in the evergreen config.


matthewdale (approved):
Thank you, I missed this! 👍

PR #175, opened 2023-09-18
DRIVERS-2715 Investigate Increased Costs of Drivers Atlas QA Usage

blink1073:
The clusters were silently failing to delete because they were trying to delete the cluster in `drivers-atlas-testing` instead of the the newly created project.

### DIFF
--- a/astrolabe/cli.py
+++ b/astrolabe/cli.py
@@ -143,7 +143,7 @@ def cli(ctx, atlas_base_url, atlas_api_username,
             timeout=http_timeout)
     else:
         admin_client = None
-    
+
     ctx.obj = ContextStore(client, admin_client)
 
     # Configure logging.
@@ -558,7 +558,7 @@ def get_logs_cmd(ctx, spec_test_file, workload_file, org_id, project_name,
     Retrieves logs for the cluster and saves them in logs.tar.gz in the
     current working directory.
     """
-    
+
     if only_on_failure:
         if os.path.exists('status'):
             with open('status') as fp:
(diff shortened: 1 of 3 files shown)


blink1073:
It looks like these failures were pre-existing in https://github.com/mongodb-labs/drivers-atlas-testing/pull/161.

astrolabe/cli.py
```diff
@@ -613,8 +615,11 @@ def delete_test_cluster(ctx, spec_test_file, workload_file, org_id, project_name
     if project:
         try:
             ctx.obj.client.groups[project.id].clusters[cluster_name].delete()
-        except AtlasApiBaseError:
-            pass
+            print(f"{msg} done.")
+        except AtlasApiBaseError as e:
+            pprint(e)
```
matthewdale, line 620: Should we re-raise this so the task fails if deleting the cluster fails? Is there something else we can do to make it more obvious that we've left an expensive resource hanging around?

astrolabe/cli.py
```diff
@@ -613,8 +615,11 @@ def delete_test_cluster(ctx, spec_test_file, workload_file, org_id, project_name
...
+    else:
+        print(f"Project {project_name} not found!")
```
matthewdale, line 622: Should we raise an exception here if we attempted to delete a cluster but couldn't find the project?

astrolabe/cli.py
blink1073: This occurs during a tear-down tastk so failing wouldn't fail the build.

astrolabe/cli.py
blink1073: Same answer as above

astrolabe/cli.py
matthewdale: Good point. We could use `post_error_fails_task: true` to make post functions fail the task.

matthewdale (approved):
Nice fix! 👍

astrolabe/cli.py
matthewdale: Created [DRIVERS-2729](https://jira.mongodb.org/browse/DRIVERS-2729) to track that suggestion.

PR #176, opened 2023-09-25
DRIVERS-2734 Ensure projects are deleted along with the cluster

blink1073:
Verified with https://spruce.mongodb.com/task/drivers_atlas_testing_tests_python__driver~pymongo_master_platform~ubuntu_18.04_runtime~python39_atlas_retryReads_move_patch_fca5526934ffbc67d485b8ec0960e843b85db9d8_6511dfcac9ec44bf4d8f36f9_23_09_25_19_30_27/logs?execution=0 that no projects remained open.

### DIFF
--- a/astrolabe/cli.py
+++ b/astrolabe/cli.py
@@ -621,6 +621,15 @@ def delete_test_cluster(ctx, spec_test_file, workload_file, org_id, project_name
     else:
         print(f"Project {project_name} not found!")
 
+    # Step-3: delete the project.
+    msg = f"Deleting project {project_name}..."
+    if project:
+        try:
+            ctx.obj.client.groups[project.id].delete()
+            print(f"{msg} done.")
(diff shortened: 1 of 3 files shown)


astrolabe/cli.py
```diff
@@ -608,19 +611,37 @@ def delete_test_cluster(ctx, spec_test_file, workload_file, org_id, project_name
...
+    msg = f"Deleting project {project_name}..."
+    if project:
+        print(msg)
+        while 1:
+            try:
+                client.groups[project.id].delete().data
+                print(f"{msg} done.")
+                break
+            except AtlasApiBaseError as e:
+                if e.error_code == 409:
```
durran, line 638: What is error code 409?

blink1073: I'll add a note, it is the error code given when the project can't be shut down because there is a running cluster.

PR #180, opened 2023-10-16
DRIVERS-2645 Use clean_old_projects to delete projects

blink1073:
Use a timestamp in the project name so the `clean_old_projects` function can remove previous projects after the fact.

### DIFF
--- a/astrolabe/cli.py
+++ b/astrolabe/cli.py
@@ -625,25 +625,6 @@ def delete_test_cluster(ctx, spec_test_file, workload_file, org_id, project_name
     else:
         print(f"Project {project_name} not found!")
 
-    # Step-3: delete the project.
-    msg = f"Deleting project {project_name}..."
-    if project:
-        print(msg)
-        while 1:
-            try:
-                client.groups[project.id].delete().data
-                print(f"{msg} done.")
-                break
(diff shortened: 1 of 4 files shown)


astrolabe/cli.py
```diff
@@ -604,17 +607,23 @@ def delete_test_cluster(ctx, spec_test_file, workload_file, org_id, project_name
...
+        client=client, org_id=org_id)
     project = cmd.get_project(
-        client=ctx.obj.client, project_name=project_name, organization_id=organization.id)
+        client=client, project_name=project_name, organization_id=organization.id)
     if project:
         try:
-            ctx.obj.client.groups[project.id].clusters[cluster_name].delete()
-        except AtlasApiBaseError:
-            pass
+            client.groups[project.id].clusters[cluster_name].delete().data
```
matthewdale, line 621: What does accessing `.data` do that's different than just calling `delete()` here?

matthewdale (approved):
Looks good! 👍

astrolabe/cli.py
blink1073: I had copied the pattern from another location.  My assumption is that that accessing the data triggers a validation.

PR #188, opened 2023-12-13
DRIVERS-2729: Make Astrolabe failure-to-cleanup more obvious

Jibola:
## Changes in this PR
Following the description on [DRIVERS-2729](https://jira.mongodb.org/browse/DRIVERS-2729), I've: 
* Added `post_error_fails_task = true` 
* Re-raise the exception on failures

## Callout
what would be the expected exception type I'd raise for when the project name is not found?

### DIFF
--- a/astrolabe/cli.py
+++ b/astrolabe/cli.py
@@ -627,8 +627,9 @@ def delete_test_cluster(ctx, spec_test_file, workload_file, org_id, project_name
             print(f"{msg} done.")
         except AtlasApiBaseError as e:
             pprint(e)
+            raise
     else:
-        print(f"Project {project_name} not found!")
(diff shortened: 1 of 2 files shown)


astrolabe/cli.py
```diff
@@ -627,8 +627,9 @@ def delete_test_cluster(ctx, spec_test_file, workload_file, org_id, project_name
             print(f"{msg} done.")
         except AtlasApiBaseError as e:
             pprint(e)
+            raise
     else:
-        print(f"Project {project_name} not found!")
+        raise AtlasClientError(f"Project {project_name} not found!")
```
blink1073, line 632: I think this error type is good

matthewdale (approved):
Looks good! 👍

PR #189, opened 2023-12-18
DRIVERS-2795 Add linters and formatters

blink1073:


blink1073:
evergreen retry

astrolabe/atlas_runner.py
```diff
@@ -131,13 +131,13 @@ def initialize(self, no_create=False):
...
-                    LOGGER.warn("Cluster was not found, will create one")
+                    LOGGER.warning("Cluster was not found, will create one")
             except AssertionError as exc:
-                LOGGER.warn(
+                LOGGER.warning(
                     "Configuration did not match: %s. Recreating the cluster" % exc
                 )
 
-        LOGGER.info("Initializing cluster {!r}".format(self.cluster_name))
+        LOGGER.info(f"Initializing cluster {self.cluster_name!r}")
```
Jibola, line 140: I personally like the cleaner look of f-strings, but I'm wondering if [we want to take on this potential drop in optimization](https://docs.python.org/3/howto/logging.html#optimization), it's not large, but I know it's still a general pythonic practice. Any particular ties to still using `.format` or `%s` replacement syntax?

Jibola (changes requested):
Left one comment on logging. Just need a response but otherwise everything else looks good to me!

astrolabe/atlas_runner.py
blink1073: Updated to use flake8-logging-format

Jibola (approved):

PR #192, opened 2024-01-17
Mark s3.puts as optional failures

Jibola:


Jibola:
Failures are no longer due to any of the functions running in the `post` stage. I think this is good to go! 
@matthewdale @blink1073

PR #201, opened 2024-03-27
PYTHON-2723 Run the PyMongo Test Class Setup Function

blink1073:


integrations/python/pymongo/workload-executor.py
```diff
@@ -33,6 +33,8 @@ def workload_runner(mongodb_uri, test_workload):
     runner = UnifiedSpecTestMixinV1()
     runner.TEST_SPEC = test_workload
     UnifiedSpecTestMixinV1.TEST_SPEC = test_workload
+    # Run the class setup function before running the instance setup function.
+    UnifiedSpecTestMixinV1.setUpClass()
```
ShaneHarvey, line 37: Does this work? I was thinking we'd need to feed mongodb_uri to the ClientContext init to get the client set up properly and avoid hitting the `client_context.require_connection` skip condition.

blink1073:
```
[2024/03/27 13:55:01.352] Ran 7 tests in 53.148s
[2024/03/27 13:55:01.352] OK
```

ShaneHarvey (approved):
LGTM, is PYTHON-2723 the correct ticket?

blink1073:
The problem was created by that ticket, yes.

PR #207, opened 2024-08-29
DRIVERS-2964 Log more information and increase wait time before idle check.

matthewdale:
[DRIVERS-2964](https://jira.mongodb.org/browse/DRIVERS-2964)

Increase the delay before checking for cluster idle status after making a config update. It's not clear why it's taking so long for the cluster status to become non-idle after issuing a config update, so the new value of 15s (up from 5s) may not suffice forever if the underlying cause of the delay gets worse. See [PRODTRIAGE-1232](https://jira.mongodb.org/browse/PRODTRIAGE-1232) for some background.

Also include the full replica set member list in the exception thrown when `assertPrimaryRegion` is unsuccessful.

### DIFF
--- a/astrolabe/atlas_runner.py
+++ b/astrolabe/atlas_runner.py
@@ -270,12 +270,14 @@ class AtlasTestCase:
                     timer.start()
                     timeout = op_spec.get("timeout", 90)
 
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
LGTM!  Let's see how this does tonight.

PR #209, opened 2024-08-30
Lower expiry threshold to 2 hours

Jibola:


astrolabe/atlas_runner.py
```diff
@@ -458,8 +453,8 @@ def __init__(
         self.persist_clusters = persist_clusters
         self.no_create = no_create
         self.workload_startup_time = workload_startup_time
-        # Hardcoded to 12 hours, can be configurable in the future
-        self.project_expiration_threshold_seconds = 12 * 60 * 60
+        # Hardcoded to 2 hours, can be configurable in the future
+        self.project_expiration_threshold_seconds = 2 * 60 * 60
```
Jibola, line 457: These are the only two lines that matter. The rest was autoformat

blink1073 (approved):
LGTM!

astrolabe/atlas_runner.py
```diff
@@ -570,8 +559,9 @@ def clean_old_projects(self, org_id):
                         else:
                             raise
                 else:
+                    LOGGER.info("Skipping deleting project %r, id: %r", project.name, project.id)
                     LOGGER.info(
-                        "Skipping deleting project %r, id: %r", project.name, project.id
+                        f"{project_timestamp=}, {current_timestamp=}, {self.project_expiration_threshold_seconds=}"  # noqa: G004
```
Jibola, line 564: @blink1073, @matthewdale  this actually snaked in, I wonder if we should still keep it as a valuable debugging datapoint

blink1073: Yes I think so

PR #210, opened 2024-09-04
DRIVERS-2964 Wait for status 'updating' after starting maintenance operations

matthewdale:


blink1073 (approved):
Thanks!  Let's merge and iterate.

PR #212, opened 2024-09-11
Use bash shell where not already in use

alcaeus:
In #211 I observed failures when running the validator. This seems to happen because Astrolabe uses `sh` to call run-mongodb.sh and subsequently call run-orchestration.sh. This PR changes scripts to use `bash` where not already in use.

Note: the PHP failure is fixed by #211, but I wanted to separate the pull requests. The .NET failure has been a consistent failure for a while.


blink1073 (approved):
Thanks!

PR #213, opened 2024-09-13
Install PHP dependencies in the project

GromNaN:
Fix PHP tests broken by https://github.com/mongodb/mongo-php-library/pull/1395

Instead of duplicating parts of the composer.json from the library, the dependencies are installed in the library directly.


alcaeus (commented):
Changes LGTM - if the validate-workload-executor tasks completes successfully we can schedule a couple of the actual tests for PHP to see if everything works correctly.

GromNaN:
Test successful for PHP.

alcaeus (approved):
Tested one test each on Atlas and Kind, both of which succeeded. Good to merge 👍

PR #214, opened 2024-09-17
DRIVERS-2964 Log cluster state for 30 minutes after assertPrimaryRegion failure.

matthewdale:
[DRIVERS-2964](https://jira.mongodb.org/browse/DRIVERS-2964)

We want to collect more information about the cluster state when the "assertPrimaryRegion" step fails, so instead of immediately stopping the job, log the cluster state every 5 seconds for the next 30 minutes (then stop the job).


blink1073 (approved):
LGTM!

PR #217, opened 2024-09-26
Rename mongo-java driver branch from master to main

jyemin:
JAVA-5608

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -678,11 +678,11 @@ axes:
           DRIVER_REVISION: "main"
           ASTROLABE_EXECUTOR_STARTUP_TIME: 45
       - id: java-master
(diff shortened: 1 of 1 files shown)


.evergreen/config.yml
```diff
@@ -678,11 +678,11 @@ axes:
           DRIVER_REVISION: "main"
           ASTROLABE_EXECUTOR_STARTUP_TIME: 45
       - id: java-master
```
jyemin, line 680: Avoided changing the id, because I couldn't prove to myself that it wouldn't have some effect on the waterfall view in Evergreen.

alcaeus (approved):
Change LGTM. I'll note that the workload validator test is failing for Java, maybe the team could take a look at that?

PR #223, opened 2024-11-04
Fix missing workload executor expansions in test runner function.

matthewdale:
Some workload executors need additional environment variables:
- Python needs `PYTHON_BIN_DIR`
- .NET needs `ASYNC` and `FRAMEWORK`
- Java needs `JAVA_HOME`

Add those to the expansions list for the "run atlas test" function.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -222,7 +222,7 @@ functions:
           ATLAS_ADMIN_API_USERNAME: ${atlas_admin_api_username}
           ATLAS_ADMIN_API_PASSWORD: ${atlas_admin_api_password}
           CLUSTER_NAME_SALT: ${build_id}
-        include_expansions_in_env: [DRIVER_DIRNAME]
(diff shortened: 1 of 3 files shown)


.evergreen/config.yml
```diff
@@ -222,7 +222,7 @@ functions:
           ATLAS_ADMIN_API_USERNAME: ${atlas_admin_api_username}
           ATLAS_ADMIN_API_PASSWORD: ${atlas_admin_api_password}
           CLUSTER_NAME_SALT: ${build_id}
-        include_expansions_in_env: [DRIVER_DIRNAME]
+        include_expansions_in_env: [DRIVER_DIRNAME, PYTHON_BIN_DIR, ASYNC, FRAMEWORK, JAVA_HOME]
```
blink1073, line 225: I think we can go back to `add_expansions_to_env: true` for the commands that use the atlas creds.

Jibola: I quite like the explicitness of `include_expansions_in_env` it ensures we know _exactly_ what's required of a script.

blink1073 (approved):
LGTM!

PR #226, opened 2024-12-19
PYTHON-4863 Fix python workload execution

blink1073:
Also fixes nightly failures we have been seeing in PyMongo.


matthewdale (approved):
Looks good! 👍 

I added this to the list of reasons we should do [DRIVERS-3044](https://jira.mongodb.org/browse/DRIVERS-3044).

